PR #17, opened 2015-01-23
[WIP] MQTT retain feature

a5anka:
Current implementation only support H2 in-memory message store.

Known issues
- Obsolete retained content is not removed from DB


a5anka:
This is done in https://github.com/wso2/andes/pull/115

PR #24, opened 2015-02-09
Remove onflight message tracker

abeykoon:


ramith:
As usual @abeykoon  bringing in the awsomeness during the week end !!! . BTW, did you break MQTT stuff or it still working?

PR #34, opened 2015-02-19
Upgrade Disruptor version to 3.3.0.wso2v1

Asitha:
- Removed ConcurrentBatchProcessor (custom batch processor) to avoid
  difficulties in future upgrades due to custom batch processors' high dependency 
  on disruptor internals. Now using EventHandler as a custom batch processor to avoid this.

- Changed MQTT disruptor ring creation to use Disruptor DSL.


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/distruptor/inbound/ConcurrentBatchEventHandler.java
```diff
@@ -0,0 +1,117 @@
+/*
+ * Copyright (c) 2014, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
```
ramith: Invalid license  Header!

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/distruptor/inbound/ConcurrentBatchEventHandler.java
```diff
@@ -0,0 +1,117 @@
...
+ */
+public class ConcurrentBatchEventHandler implements EventHandler<InboundEvent> {
+
+    private static Log log = LogFactory.getLog(ConcurrentBatchEventHandler.class);
+
+
+    private final BatchEventHandler eventHandler;
+    private final long turn;
+    private final int groupCount;
+    private int batchSize;
```
ramith: Shall we explain these variables ?!!! so others can understand the magic you guys brought in!

ramith:
wow... we finally got the disruptor upgraded!!!

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/distruptor/inbound/ConcurrentBatchEventHandler.java
Asitha: ConcurrentBatchProcessor class was renamed to ConcurrentBatchEventHandler after having a discussion [1] at architecture@wso2.org. Header is hence kept as 2014.

[1] Batched content chunk reading for WSO2 MB through Disruptor

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/distruptor/inbound/ConcurrentBatchEventHandler.java
Asitha: Variable comments added with commit https://github.com/Asitha/andes/commit/5ddfe4d9dcb44bcf1edd7f5fe769443c17d45cc6

PR #42, opened 2015-02-25
[MB-958] Fixing the problems with the build of andes with oracle jdk 1.8

Megala21:
Compatibility issues fixed.


Megala21:
Public jira  - https://wso2.org/jira/browse/MB-958

PR #50, opened 2015-02-27
Fix overlapping slots

hastef88:
**Issue**

In a clustered setup, according our current slot coordination flow, if 2 slots sent by nodeA and nodeB overlap, we merge it into one slot and assign it to 1 node (e.g. nodeA). This could cause issues since if nodeB is still writing messages within that slot by itself, nodeA will have no idea about it and will try to delete the slot, based on the deliveredMessageCount = nodeA's messages within the slot range. 

The fix handles this in 3 approaches : 

**Safe Zone calculation** 

The min(last_generated_message_id) of all nodes is derived from time to time. (15 seconds) and is used to decide if a slot can be deleted. If a node is inactive (does not have publishers), it will send a heartbeat message ID generated from time to time, which is used for the calculation.

During SlotManager.deleteSlot() , the slot is deleted only if the slot.endMessageID is less than the safeZone. 
If the safezone is past the slot.endMessageID, that means that all nodes have moved past processing any messages possibly processed within that slot.

**Overlapping slot detection**

A new hazelcast map "overlappedSlotMap" was added to maintain all overlapping slots per node and destination (queue). 

Whenever the SlotManager.updateMessageID thrift service is called by a node (sending a slot), the sent slot range will be compared against all ASSIGNED slots to find if it overlaps with any. If it does, the respective assigned slots are marked as "OVERLAPPED", and stored into the overlappedSlotMap. 

Whenever nodeA requests for a slot using the SlotManager.getSlot() service, coordinator will first check if there are any overlapping slots, and handover the firstmost overlapping slot to that node for processing. This will ensure that nodeA gets a chance to capture any messages sent to other nodes within that slot. 

**Slot state model**

A state model was introduced for all slots so that we can ensure that their lifecycle happens as we intend to. This was done using an enum implementation (refer SlotState class) 

new Slot() -> CREATED
CREATED -> ASSIGNED 
ASSIGNED -> OVERLAPPED or RETURNED or DELETED
OVERLAPPED -> ASSIGNED or OVERLAPPED
RETURNED -> ASSIGNED or OVERLAPPED
DELETED -> slot.destroy()

More information on mail : [ARCHITECTURE][WSO2 Message Broker] Slot Based Message Distribution and Slot Lifecycle


ramith:
Nice PR description.

modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/gen/SlotManagementService.java
```diff
@@ -1,2958 +0,0 @@
-/*
```
hastef88: This is a file generated using thrift-gen

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/Andes.java
```diff
@@ -107,9 +120,23 @@ void initialise(SubscriptionStore subscriptionStore,
         this.subscriptionManager = subscriptionManager;
 
         inboundEventManager = InboundEventManagerFactory.createEventManager(subscriptionStore, messagingEngine);
+
         log.info("Andes API initialised.");
     }
 
+    public void startSafeZoneAnalysisWorker() {
```
ramith: safe zone for what ?

hastef88: Added method level comments

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/Slot.java
```diff
@@ -101,6 +130,17 @@ public boolean isSlotActive() {
...
     }
 
+    public boolean isAnOverlappingSlot() {
+        return isAnOverlappingSlot;
+    }
+
+    public void setAnOverlappingSlot(boolean isAnOverlappingSlot) {
+        this.isAnOverlappingSlot = isAnOverlappingSlot;
+        if(isAnOverlappingSlot) {
+            addState(SlotState.OVERLAPPED);
```
pamod: Would it be meaningful to add debug logs here ? Mean to have a way to track each of the states represented by the slot

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManager.java
```diff
@@ -57,27 +61,59 @@
...
             queueToLastAssignedIDMap = hazelcastAgent.getLastAssignedIDMap();
+            nodeToLastPublishedIDMap = hazelcastAgent.getLastPublishedIDMap();
             slotAssignmentMap = hazelcastAgent.getSlotAssignmentMap();
 
+            nodeInformedSlotDeletionSafeZones = new HashMap<String, Long>();
+
+            //start a thread to calculate slot delete safe zone
+            //TODO: use a common  thread pool for tasks like this?
+            slotDeleteSafeZoneCalc = new SlotDeleteSafeZoneCalc(SAFE_ZONE_EVALUATION_INTERVAL);
+            new Thread(slotDeleteSafeZoneCalc).start();
```
pamod: As mentioned in the comment great if we can get it into a thread-pool and then name the pool as well. Would come in handy when debugging

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManager.java
```diff
@@ -96,50 +132,43 @@ public static SlotManager getInstance() {
...
-            } else {
-                if (log.isDebugEnabled()) {
-                    log.debug("Slot Manager - returns empty slot for the queue: " + queueName);
-                }
+        }
+
+        if (null != slotToBeAssigned) {
+            updateSlotAssignmentMap(queueName, slotToBeAssigned, nodeId);
+            if (log.isDebugEnabled()) {
+                log.info("Assigning slot for node : " + nodeId + " ||| " + slotToBeAssigned);
```
pamod: We need to change this to a debug log ?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/Slot.java
hastef88: Logging is done at the higher level when marking the slot as overlapped. (We have more context there for logging)

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManager.java
hastef88: Added with commit : https://github.com/abeykoon/andes/commit/f81869389e6d660779cbd5357d7e77c9f6fb52fc

Asitha:
Can you pls explain what happens when a node takes an overlapped slot? (From the above explanation overlapped slot can be taken from any node)

hastef88:
not exactly. An overlapped slot here is an already assigned slot, meaning it has been mapped to a specific node ID. So when that node requests another slot, the earliest overlapped slot of that node will be handed over. This way we can ensure that a slot is continuously processed by the same node once it is assigned to it.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/Slot.java
```diff
@@ -18,19 +18,28 @@
...
+import org.apache.commons.lang.StringUtils;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
 import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * This class stores all the data related to a slot
```
Asitha: It would be nice if we explain what is a slot? A brief explanation of how we use it. (BTW we have Slots in Disruptor too.)

Asitha:
Great! That sounds right.

I've slightly updated your description as follows to convey that idea. 
"Whenever nodeA requests for a slot using the SlotManager.getSlot() service, coordinator will first check if there are any overlapping slots for nodeA (requesting node) in already assigned slots, and handover the firstmost overlapping slot to that node for processing."

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/Slot.java
```diff
@@ -52,6 +61,16 @@
     private boolean isSlotActive;
 
     /**
+     * Indicates whether the slot is a fresh one or an overlapped one
+     */
+    private boolean isAnOverlappingSlot;
```
Asitha: Once overlapped we keep this boolean true till it gets deleted? 
For instance when we change state from say, overlapped -> assigned we still need to check its ovelappin. Is that why we keep this information separate rather than using SlotState for this purpose?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/Slot.java
```diff
@@ -18,19 +18,28 @@
 
 package org.wso2.andes.kernel.slot;
```
Asitha: Can't we have unit tests for this class? IMO some change here will lead to subtle changes in functionality which might be really difficult to debug.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/Slot.java
```diff
@@ -109,6 +149,64 @@ public void setDestinationOfMessagesInSlot(String destinationOfMessagesInSlot) {
...
+     * Check if state going to be added is valid considering it as the next
+     * transition compared to current latest state.
+     * @param state state to be transferred
+     */
+    public boolean addState(SlotState state) {
+
+        boolean isValidTransition = false;
+
+        if(slotStates.isEmpty()) {
+            if(SlotState.CREATED.equals(state)) {
```
Asitha: state.equals(SlotState.CREATED) is more readable I guess?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/Slot.java
```diff
@@ -109,6 +149,64 @@ public void setDestinationOfMessagesInSlot(String destinationOfMessagesInSlot) {
...
+     * Convert Slot state list to a string
+     * @return Encoded string
+     */
+    public String encodeSlotStates() {
+        String encodedString;
+        StringBuilder builder = new StringBuilder();
+        for (SlotState slotState : slotStates) {
+            builder.append(slotState.getCode()).append("%");
+        }
+        encodedString = builder.toString();
```
Asitha: what about just return `builder.toString()`? Get rid of the `encodedString` variable

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/Slot.java
```diff
@@ -109,6 +149,64 @@ public void setDestinationOfMessagesInSlot(String destinationOfMessagesInSlot) {
...
+        slotStates.clear();
+        for (String state : states) {
+            int code = Integer.parseInt(state);
+            slotStates.add(SlotState.parseSlotState(code));
+        }
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
```
Asitha: WSO2 standard if block?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/Slot.java
```diff
@@ -149,9 +243,11 @@ public String getId() {
         return storageQueueName + "|" + startMessageId + "-" + endMessageId;
     }
 
+    @SuppressWarnings("NullableProblems")
     @Override
     public int compareTo(Slot other) {
-        if(this.getStartMessageId() == other.getStartMessageId()) {
+        if ((this.getStartMessageId() == other.getStartMessageId()) && (this.getEndMessageId() == other
```
Asitha: What about we rearrange this similar to following. Would it be more readable?

```java
 if ( ( this.getStartMessageId() == other.getStartMessageId() ) && 
      ( this.getEndMessageId() == other.getEndMessageId() ) &&
      this.getStorageQueueName().equals(other.getStorageQueueName() )) { 
```

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeleteSafeZoneCalc.java
```diff
@@ -0,0 +1,206 @@
...
+                    //If messages are not published, each node will send a messageID giving
+                    // assurance that next message id it would generate will be beyond a certain
+                    // number
+                    Long nodeInformedSafeZone = nodeInformedSafeZones.get(nodeID);
+
+                    /**
+                     * if no new messages are published and no new slot assignment happened
+                     * node informed value can be bigger. We need to accept that to keep the
+                     * safe zone moving
+                     */
```
Asitha: Do we need Doc comments here?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeleteSafeZoneCalc.java
```diff
@@ -0,0 +1,206 @@
...
+        while (running) {
+            if (isLive) {
+                Set<String> nodesWithPublishedMessages = SlotManager.getInstance().getMessagePublishedNodes();
+                Map<String, Long> nodeInformedSafeZones =
+                        SlotManager.getInstance().getNodeInformedSlotDeletionSafeZones();
+
+                /** calculate safe zone (minimum value of messageIDs published so far to the
+                 * cluster by each node)
+                 */
+                long globalSafeZoneVal = Long.MAX_VALUE;
```
Asitha: Java Doc comments?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeleteSafeZoneCalc.java
```diff
@@ -0,0 +1,206 @@
...
+                try {
+                    Thread.sleep(seekInterval);
+                } catch (InterruptedException e) {
+                    //silently ignore
+                }
+            } else {
+                try {
+                    Thread.sleep(15 * 1000);
+                } catch (InterruptedException e) {
+                    //silently ignore
```
Asitha: Do we need to ignore a thread interrupt here? Can't we have something like
`thread.currentthread().interrupt()`

Anyway can't we have this whole thing as a sheduled task. Do we need to go in an infinite loop updating the safe zone and sleeping. This is not in critical path right? Plus wouldn't it make this Thread.sleep() stuff go away?

Asitha:
If we have ovelapped slots belonging to multiple nodes? Is the slot divided to multiple parts or how it is handled?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/Slot.java
hastef88: Yes this is for the convenience factor. The Slotstates are maintained within the object as a list, and to refer the list to find out the overlapping slots would be unnecessary. the states are maintained to validate the transition from one stage to the other.

Asitha: If the order of states saved (ArrayList) doesn't matter we can use a HashSet for States and use HashSet#contains to have the same behaviour. Just an Idea.

hastef88:
In case a slot is overlapping with multiple slots on different nodes, all such slots will be marked as overlapped along with their association to that node. So they will be fetched and processed by the respective nodes as expected.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/Slot.java
hastef88: Yes, that should work given that only overlapped slots are given repetitively. Will check on it. Good suggestion.

PR #59, opened 2015-03-09
A fix with reference to https://wso2.org/jira/browse/MB-927. A more mean...

pamod:
...ingful error message was added to indicate the fow control wait timeout.


modules/andes-core/client/src/main/java/org/wso2/andes/client/AMQSession.java
```diff
@@ -477,11 +477,11 @@ public void aboveThreshold(int currentValue)
                                                                  // was False
                                                                  if (!_suspendState.getAndSet(true))
                                                                  {
```
Asitha: Code style from Qpid code.

PR #68, opened 2015-03-12
Fix Jiras 964 and 970

hastef88:
Handling exceptions occurring during server shutdown. Utilize Carbon ShutdownEventListener to close any isolated worker threads (e.g. SlotManager).

Handle HazelcastInstanceNotActiveException

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
@@ -68,7 +68,7 @@ public class AndesKernelBoot {
     /**
      * This is used by independent worker threads to identify if the kernel is performing shutdown operations.
      */
-    private static boolean isKernelShuttingDown;
+    private static boolean isKernelShuttingDown = false;
 
     /**
      * This will boot up all the components in Andes kernel and bring the server to working state
@@ -188,21 +188,8 @@ public class AndesKernelBoot {
         // Set flag so independent threads can act accordingly
         isKernelShuttingDown = true;
 
-        stopMessaging();
-
-        //Also add to Un register any MBeans at this point as needed
-
-        cleanUpAndNotifyCluster();
-        stopHouseKeepingThreads();
-
-        stopThriftServer();
-
+        // Trigger Shutdown Event
         Andes.getInstance().shutDown();
-
-        ApplicationRegistry.remove();
-
-        //close stores
-        AndesContext.getInstance().getAndesContextStore().close();
 
     }
 
@@ -306,9 +293,8 @@ public class AndesKernelBoot {
      */
     private static void cleanUpAndNotifyCluster() throws AndesException {
         //at the shutDown close all localSubscriptions and notify cluster
-        log.info("Closing all local subscriptions existing...");
-        ClusterResourceHolder.getInstance().getSubscriptionManager()
-                             .closeAllLocalSubscriptionsOfNode();
+
+        ClusterResourceHolder.getInstance().getSubscriptionManager().closeAllLocalSubscriptionsOfNode();
         // notify cluster this MB node is shutting down. For other nodes to do recovery tasks
         ClusterResourceHolder.getInstance().getClusterManager().shutDownMyNode();
     }
@@ -332,7 +318,7 @@ public class AndesKernelBoot {
      *
      * @throws Exception
      */
-    private static void stopHouseKeepingThreads() {
+    public static void stopHouseKeepingThreads() {
         log.info("Stop syncing exchanges, queues, bindings and subscriptions...");
         int threadTerminationTimePerod = 20; // seconds
         try {
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
@@ -20,6 +20,7 @@ package org.wso2.andes.kernel;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.wso2.andes.amqp.QpidAMQPBridge;
 import org.wso2.andes.configuration.AndesConfigurationManager;
 import org.wso2.andes.configuration.StoreConfiguration;
 import org.wso2.andes.configuration.enums.AndesConfiguration;
@@ -31,6 +32,7 @@ import org.wso2.andes.server.cluster.coordination.hazelcast.HazelcastAgent;
 import org.wso2.andes.server.information.management.MessageStatusInformationMBean;
 import org.wso2.andes.server.information.management.SubscriptionManagementInformationMBean;
 import org.wso2.andes.server.queue.DLCQueueUtils;
+import org.wso2.andes.server.registry.ApplicationRegistry;
 import org.wso2.andes.thrift.MBThriftServer;
 import org.wso2.andes.server.virtualhost.VirtualHost;
 import org.wso2.andes.server.virtualhost.VirtualHostConfigSynchronizer;
@@ -62,10 +64,16 @@ public class AndesKernelBoot {
     private static AndesContextStore contextStore;
 
     /**
+     * This is used by independent worker threads to identify if the kernel is performing shutdown operations.
+     */
+    private static boolean isKernelShuttingDown;
+
+    /**
      * This will boot up all the components in Andes kernel and bring the server to working state
      */
     public static void bootAndesKernel() throws Exception {
 
+        isKernelShuttingDown = false;
         //loadConfigurations - done from outside
         //startAndesStores - done from outside
         int threadPoolCount = 1;
@@ -171,19 +179,29 @@ public class AndesKernelBoot {
     /**
      * This will trigger graceful shutdown of andes broker
      *
-     * @throws Exception
+     * @throws AndesException
      */
-    public static void shutDownAndesKernel() throws Exception {
+    public static void shutDownAndesKernel() throws AndesException {
+
+        // Set flag so independent threads can act accordingly
+        isKernelShuttingDown = true;
 
         stopMessaging();
-        stopThriftServer();
-        unregisterMBeans();
+
+        //Also add to Un register any MBeans at this point as needed
+
         cleanUpAndNotifyCluster();
         stopHouseKeepingThreads();
-        stopAndesComponents();
+
+        stopThriftServer();
+
+        Andes.getInstance().shutDown();
+
+        ApplicationRegistry.remove();
+
         //close stores
         AndesContext.getInstance().getAndesContextStore().close();
-        Andes.getInstance().shutDown();
+
     }
 
     /**
@@ -282,9 +300,9 @@ public class AndesKernelBoot {
     /**
      * clean up broker states and notify the cluster
      *
-     * @throws Exception
+     * @throws AndesException
      */
-    private static void cleanUpAndNotifyCluster() throws Exception {
+    private static void cleanUpAndNotifyCluster() throws AndesException {
         //at the shutDown close all localSubscriptions and notify cluster
         log.info("Closing all local subscriptions existing...");
         ClusterResourceHolder.getInstance().getSubscriptionManager()
@@ -312,7 +330,7 @@ public class AndesKernelBoot {
      *
      * @throws Exception
      */
-    private static void stopHouseKeepingThreads() throws Exception {
+    private static void stopHouseKeepingThreads() {
         log.info("Stop syncing exchanges, queues, bindings and subscriptions...");
         int threadTerminationTimePerod = 20; // seconds
         try {
@@ -320,9 +338,9 @@ public class AndesKernelBoot {
             andesRecoveryTaskScheduler
                     .awaitTermination(threadTerminationTimePerod, TimeUnit.SECONDS);
         } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
             andesRecoveryTaskScheduler.shutdownNow();
             log.warn("Recovery task scheduler is forcefully shutdown.");
-            throw e;
         }
 
     }
@@ -346,15 +364,6 @@ public class AndesKernelBoot {
         MessageStatusInformationMBean messageStatusInformationMBean = new
                 MessageStatusInformationMBean();
         messageStatusInformationMBean.register();
-    }
-
-    /**
-     * unregister any MBeans registered
-     *
-     * @throws Exception
-     */
-    private static void unregisterMBeans() throws Exception {
-
     }
 
     /**
@@ -387,15 +396,6 @@ public class AndesKernelBoot {
     }
 
     /**
-     * Stop andes components
-     *
-     * @throws Exception
-     */
-    private static void stopAndesComponents() throws Exception {
-
-    }
-
-    /**
      * Start accepting and delivering messages
      *
      * @throws Exception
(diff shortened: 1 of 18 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
```diff
@@ -62,10 +64,16 @@
...
 
     /**
+     * This is used by independent worker threads to identify if the kernel is performing shutdown operations.
+     */
+    private static boolean isKernelShuttingDown;
+
+    /**
      * This will boot up all the components in Andes kernel and bring the server to working state
      */
     public static void bootAndesKernel() throws Exception {
```
hemikak: Is it necessary to throw "Exception" ? is there no particular exception ? :)

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
```diff
@@ -450,4 +451,8 @@ private static void createSuperTenantDLC() throws AndesException {
             throw new AndesException("Error getting super tenant username", e);
         }
     }
+
+    public static boolean isKernelShuttingDown() {
```
hemikak: JavaDoc missing.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessagingEngine.java
```diff
@@ -627,13 +631,15 @@ public void stopMessageDelivery() {
         if (MBThriftClient.isReconnectingStarted()) {
             MBThriftClient.setReconnectingFlag(false);
         }
-        log.info("Stopping Disruptor writing messages to store.");
+        SlotMessageCounter.getInstance().stop();
+        log.info("Stopping Disruptor workers from delivering messages to store.");
     }
 
     public void close() throws InterruptedException {
```
hemikak: JavaDoc missing.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManagerClusterMode.java
```diff
@@ -555,77 +560,81 @@ public boolean deleteSlot(String queueName, Slot emptySlot, String nodeId) {
...
 
-            if(log.isDebugEnabled()) {
-                log.debug("Cleared assigned slots of queue " + queueName + " Assigned to node " +
-                        nodeId);
-            }
+                //Get assigned slots from Hazelcast, delete all belonging to queue
+                //and set back
+                HashmapStringTreeSetWrapper wrapper = slotAssignmentMap.get(nodeId);
+                HashMap<String, TreeSet<Slot>> queueToSlotMap = null;
+                if (wrapper != null) {
```
hemikak: flip ? 
null != wrapper

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
```diff
@@ -62,10 +66,16 @@
     private static AndesContextStore contextStore;
 
     /**
+     * This is used by independent worker threads to identify if the kernel is performing shutdown operations.
+     */
+    private static boolean isKernelShuttingDown;
```
sajinidesilva: I think its ok to initialize this variable with default value false

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManagerClusterMode.java
```diff
@@ -338,93 +338,98 @@ private void updateSlotAssignmentMap(String queueName, Slot allocatedSlot, Strin
...
-                            lastMessageIdInTheSlot + " to slotIDMap");
+                    //record last published message ID
+                    nodeToLastPublishedIDMap.set(nodeId, lastMessageIdInTheSlot);
                 }
-
-                //record last published message ID
-                nodeToLastPublishedIDMap.set(nodeId, lastMessageIdInTheSlot);
             }
+        } else {
+            log.warn("The server is shutting down and therefore the cluster is inactive. Therefore updating the slot " +
```
sajinidesilva: Here you have to check whether the server is shutting down I guess. Its better if you can catch the HazelcastInstanceNotActiveException and if the server is shutting down ignore the exception. 

} catch (HazelcastInstanceNotActiveException e) {
                           if(!serverlsShuttingDown){
                                 log.error("Hazelcast instance is not active and therefore cannot update the slot manager", e);
                           }
                        // Ignore this exception if the server is shutting down
                    }

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeliveryWorker.java
```diff
@@ -87,6 +88,13 @@ public void run() {
                         if (messageFlusher.getMessageDeliveryInfo(destinationOfMessagesInQueue)
                                 .isMessageBufferFull()) {
 
+                            //If hazelcast is inactive at this moment, we cannot proceed further with the clustered
+                            // data collections. therefore warning and skipping.
+                            if (AndesContext.getInstance().isClusteringEnabled() && !HazelcastAgent.getInstance().isActive()) {
```
sajinidesilva: It not good to check it here like that. Because hazelcast instance can go down just after this line and therefore exception can occur. Better to use a try catch and catch the HazelcastInstanceNotActiveException and ignore the exception is the server is shutting down. Otherwise print a error log.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManagerClusterMode.java
```diff
@@ -670,58 +679,62 @@ public long updateAndReturnSlotDeleteSafeZone(String nodeID, long safeZoneOfNode
      */
     public void clearAllActiveSlotRelationsToQueue(String queueName) {
 
-        if(log.isDebugEnabled()) {
-            log.debug("Clearing all slots for queue " + queueName);
-        }
+        if (HazelcastAgent.getInstance().isActive()) {
+            if(log.isDebugEnabled()) {
+                log.debug("Clearing all slots for queue " + queueName);
```
sajinidesilva: See my above comments

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManagerClusterMode.java
```diff
@@ -670,58 +679,62 @@ public long updateAndReturnSlotDeleteSafeZone(String nodeID, long safeZoneOfNode
...
-        }
+            if (null != slotIDMap) {
+                slotIDMap.remove(queueName);
+            }
 
-        // Clear slots assigned to the queue along with overlapped slots
-        if (AndesContext.getInstance().isClusteringEnabled()) {
-            String nodeId = HazelcastAgent.getInstance().getNodeId();
+            // Clear slots assigned to the queue along with overlapped slots
+            if (AndesContext.getInstance().isClusteringEnabled()) {
```
sajinidesilva: You don't need to check is clustering enabled here because this class will be initiated only in cluster mode. Therefore all the clustering enabled checking in this class are not necessary and they should be removed.

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/coordination/hazelcast/HazelcastAgent.java
```diff
@@ -362,38 +362,54 @@ public int getIndexOfLocalNode() {
...
-        } catch (Exception e) {
-            log.error("Error while sending exchange change notification" + clusterNotification.getEncodedObjectAsString(), e);
-            throw new AndesException("Error while sending exchange change notification" + clusterNotification.getEncodedObjectAsString(), e);
+        if (isActive()) {
+            if (log.isDebugEnabled()) {
+                log.debug("Sending GOSSIP: " + clusterNotification.getDescription());
+            }
+            try {
+                this.exchangeChangeNotifierChannel.publish(clusterNotification);
+            } catch (Exception e) {
```
sajinidesilva: Why are we catching Exception here. It may includes runtime exceptions as well. Is there a special reason to do that?

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/coordination/hazelcast/HazelcastAgent.java
```diff
@@ -480,4 +496,16 @@ public void releaseInitializationLock() {
         }
     }
 
+    /**
+     * Method to check if the hazelcast instance has shutdown.
+     * @return boolean
+     */
+    public boolean isActive() {
+        if (null != hazelcastInstance) {
+            return hazelcastInstance.getLifecycleService().isRunning();
```
sajinidesilva: This method is not necessary if we catch the HazelcastInstanceNotActiveException. WDYT?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
```diff
@@ -171,19 +181,29 @@ public static void setVirtualHost(VirtualHost defaultVirtualHost) {
...
      * This will trigger graceful shutdown of andes broker
      *
-     * @throws Exception
+     * @throws AndesException
      */
-    public static void shutDownAndesKernel() throws Exception {
+    public static void shutDownAndesKernel() throws AndesException {
+
+        // Set flag so independent threads can act accordingly
+        isKernelShuttingDown = true;
```
sajinidesilva: Can we set this value using the shutdown listener in kernel? Setting the value here will be too late.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessagingEngine.java
```diff
@@ -644,6 +654,11 @@ public void close() throws InterruptedException {
             log.warn("Content remover task forcefully shutdown.");
             throw e;
         }
+
+        //Stop Slot manager in coordinator
+        if (AndesContext.getInstance().isClusteringEnabled() && AndesContext.getInstance().getClusteringAgent().isCoordinator()) {
+            ClusterResourceHolder.getInstance().getClusterManager().getSlotManager().shutDownSlotManager();
```
sajinidesilva: What is done by shutting down the slot manager? Can you please explain?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
hastef88: Fixed :) used specific exceptions.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
hastef88: Since this is a de-facto getter method, the intention is to refer to the member variable comment for information.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessagingEngine.java
hastef88: Added

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManagerClusterMode.java
hastef88: done

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
hastef88: done

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManagerClusterMode.java
hastef88: Fixed as suggested. The error will only be printed if the server is not in shutting_down state. If it is, the log is only printed in debug mode.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeliveryWorker.java
hastef88: Removed and this is now covered by catching the exceptions in Slotmanager

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManagerClusterMode.java
hastef88: Removed all clustering enabled checks from SlotManagerClusterMode

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/coordination/hazelcast/HazelcastAgent.java
hastef88: Agreed. Fixed as suggested. The error will only be printed if the server is not in shutting_down state. If it is, the log is only printed in debug mode.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
hastef88: Shutdown listener is not intended to invoke any custom logic before or after shutting down the andes kernel. (apart from unergistering the QpidService from osgi.) So setting the value here is timely.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessagingEngine.java
hastef88: As at now, this method will close the SlotDeletionSafeZone calculating worker threads. if needed we can add any similar workers here in future.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/distruptor/inbound/InboundKernelOpsEvent.java
```diff
@@ -59,7 +63,7 @@ public void updateState() throws AndesException {
                 stopMessageExpirationWorker();
                 break;
             case SHUTDOWN_MESSAGING_ENGINE_EVENT:
-                shutdownMessagingEngine();
+                gracefulShutdown();
```
ramith: shouldn't it be like : gracefullyShutdown()...!! (just a optional very trivial change)

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManagerClusterMode.java
```diff
@@ -95,26 +100,24 @@
...
+        nodeInformedSlotDeletionSafeZones = new HashMap<String, Long>();
 
-            //start a thread to calculate slot delete safe zone
-            //TODO: use a common  thread pool for tasks like this?
-            slotDeleteSafeZoneCalc = new SlotDeleteSafeZoneCalc(SAFE_ZONE_EVALUATION_INTERVAL);
-            new Thread(slotDeleteSafeZoneCalc).start();
+        //start a thread to calculate slot delete safe zone
+        //TODO: use a common  thread pool for tasks like this?
+        slotDeleteSafeZoneCalc = new SlotDeleteSafeZoneCalc(SAFE_ZONE_EVALUATION_INTERVAL);
+        new Thread(slotDeleteSafeZoneCalc).start();
```
ramith: So we start a thread. when are we stopping this thread...?
(during shutdown?)

hastef88: This is done at the shutDownSlotManager method which is invoked as part of server shutdown.

PR #71, opened 2015-03-19
Hybrid Subscription store (Review only do not merge)

akalankapagoda:
This has the new Subscription Store with hybrid model for handling subscriptions.
Handle direct subscriptions from within the subscription store and wildcard subscriptions from a SubscriptionHandler.


modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPUtils.java
```diff
@@ -418,17 +420,33 @@ public static boolean isTargetQueueBoundByMatchingToRoutingKey(String queueBound
...
+    /**
+     * Checks whether a given subscription is a wildcard subscription.
+     *
+     * @param subscribedDestination The destination string subscriber subscribed to
+     * @return is this a wild card subscription
+     */
+    public static boolean isWildCardSubscription(String subscribedDestination) {
+        boolean isWildCard = false;
+
+        if (subscribedDestination.contains(TOPIC_AND_CHILDREN_WILDCARD) || subscribedDestination.contains(IMMEDIATE_CHILDREN_WILDCARD)) {
```
a5anka: Shouldn't we flip to avoid null pointer exception?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/distruptor/inbound/MessagePreProcessor.java
```diff
@@ -112,11 +113,19 @@ private void handleTopicRoutine(InboundEvent event, AndesMessage message, AndesC
...
-            // TODO: This call is O(n2). In critical path. Need to improve
-            subscriptionList = subscriptionStore.getClusterSubscribersForDestination(messageRoutingKey, true);
+            // Get subscription list according to the message type
+            AndesSubscription.SubscriptionType subscriptionType;
+
+            if (MessageMetaDataType.META_DATA_MQTT == message.getMetadata().getMetaDataType()) {
+                subscriptionType = AndesSubscription.SubscriptionType.MQTT;
+            } else {
+                subscriptionType = AndesSubscription.SubscriptionType.AMQP;
+            }
```
Asitha: What if we avoid this **if** block. We can have the same enum for `SubscriptionType`  and `MessageMetaDataType`
Then we can assign without an if block. Reduce changes we have to do when a new protocol is introduced as well.
Just and idea.

modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/ClusterSubscriptionBitMapHandler.java
```diff
@@ -0,0 +1,507 @@
...
+    public void updateWildCardSubscription(AndesSubscription updatedSubscription) {
+        int index = wildCardSubscriptionList.indexOf(updatedSubscription);
+
+        if (index > -1) {
+            // Need to preserve order, hence adding to the same index
+            wildCardSubscriptionList.remove(index);
+            wildCardSubscriptionList.add(index, updatedSubscription);
+        } else {
+            log.warn("Subscription is not found to update");
+            // TODO:Warn
```
Asitha: Do we still need this TODO?

modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/ClusterSubscriptionBitMapHandler.java
```diff
@@ -0,0 +1,507 @@
...
+            int nextSetbitIndex = andBitSet.nextSetBit(0);
+
+            while (nextSetbitIndex > -1) {
+                subscriptions.add(wildCardSubscriptionList.get(nextSetbitIndex));
+                nextSetbitIndex = andBitSet.nextSetBit(nextSetbitIndex + 1);
+            }
+
+        } else {
+            log.warn("Subscription destination is empty");
+            // TODO:Warn
```
Asitha: Do we still need this TODO?

Asitha:
:+1:
Do we have any numbers?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/distruptor/inbound/MessagePreProcessor.java
akalankapagoda: Yes, that will be good. Will create a common enum for protocol type.

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPUtils.java
akalankapagoda: yes. will flip.

akalankapagoda:
Have done some testing with about 100 subscriptions and 1000 messages per each. But could not find any significant change in TPS. Need to increase the number of subscribers to above 1000 and see.

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPUtils.java
akalankapagoda: nop. This is for 'contains' not 'equals'.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/distruptor/inbound/MessagePreProcessor.java
akalankapagoda: This was not possible since there are many metadata types for AMQP. There are three meta data types in total while there are only two subscription types.

akalankapagoda:
Will merge as a new pull request.

PR #74, opened 2015-03-23
Fixed MB-985

sajinidesilva:


sajinidesilva:
Tracked by https://wso2.org/jira/browse/MB-985

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManagerClusterMode.java
```diff
@@ -145,10 +145,10 @@ public Slot getSlot(String queueName, String nodeId) {
         String lockKey = queueName + SlotManagerClusterMode.class;
 
         synchronized (lockKey.intern()) {
-            slotToBeAssigned = getOverlappedSlot(nodeId, queueName);
+            slotToBeAssigned = getUnassignedSlot(queueName);
```
hastef88: Is it right to fetch an unassigned slot before checking overlapped ones ? the reasoning here was to ensure that overlapped slots are given priority when giving a slot. (So that before reaching an unassigned slot, overlaps can be handled)

sajinidesilva: We thought that unassigned slots should be given priory since they are the oldest messages

PR #79, opened 2015-03-24
Generate slots from begining in standalone mode when server starts up

sajinidesilva:


sajinidesilva:
Tracked by https://wso2.org/jira/browse/MB-995

PR #85, opened 2015-03-27
Allowed empty messages to be received to the subscriber.

hemikak:
When an empty https://docs.oracle.com/javaee/6/api/javax/jms/TextMessage.html is published, the "ContentCacheCreator" is giving a null exception.


ramith:
SO what does the spec says?

hemikak:
The spec does not specify the case when the message body is empty.  The spec mentions the following - "Messages carry a content, which consists of a content header, holding a set of properties, and a content body, holding an opaque block of binary data.". I checked against ActiveMQ as well, and it is possible to publish a TextMessage with empty body.
Even though the body is empty, still the user can add a set of properties to the message and publish.

ramith:
Good to know! :)

PR #103, opened 2015-04-21
[MB-703] - Deactivating durable subscription when node is killed.

hemikak:
The PR contains a fix for durable topics when it comes to fail-over. When a node is killed, the durable subscription is kept active by "hasExternalSubscriber" being true, therefore when the fail-over subscriber attempts to reconnect, it mentions that the subscriber is active.

JIRA - https://wso2.org/jira/browse/MB-703

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesSubscriptionManager.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesSubscriptionManager.java
@@ -30,7 +30,6 @@ import org.wso2.andes.subscription.BasicSubscription;
 import org.wso2.andes.subscription.SubscriptionStore;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesSubscriptionManager.java
```diff
@@ -100,26 +99,31 @@ public void addSubscription(LocalSubscription localSubscription) throws AndesExc
...
      * @throws AndesException
      */
     public void closeAllClusterSubscriptionsOfNode(String nodeID) throws AndesException {
 
-        Set<AndesSubscription> activeSubscriptions = subscriptionStore.getActiveClusterSubscribersForNode(nodeID, true);
-        activeSubscriptions.addAll(subscriptionStore.getActiveClusterSubscribersForNode(nodeID, false));
+        Set<AndesSubscription> activeSubscriptions = subscriptionStore
+                                                .getActiveClusterSubscribersForNode(nodeID, true);
+        activeSubscriptions.addAll(subscriptionStore.getActiveClusterSubscribersForNode(nodeID,
+                                                                                            false));
```
Asitha: Shall we move the two arguments of the method to a single line?

hemikak: Will do.

PR #109, opened 2015-04-23
Publisher Acknowledgement implementation

Asitha:
- Andes pub ack support added.
- MQTT pub ack integrated with Andes pub ack model

JIRA https://wso2.org/jira/browse/MB-1022


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/DisablePubAckImpl.java
```diff
@@ -0,0 +1,27 @@
+package org.wso2.andes.kernel;
+
+import org.wso2.andes.kernel.distruptor.inbound.PubAckHandler;
+
+/**
+ * Drops any request to send publisher acknowledgements.
+ */
+public class DisablePubAckImpl implements PubAckHandler{
```
ramith: need a better comment than this...
explain why we need this?

PR #119, opened 2015-05-13
[MB-875] Fix to provide a hostname in broker.xml

hemikak:
Used "InetAddress.getByName(bindAddressFromBrokerOptions);" instead of "InetAddress.getByAddress(parseIP(bindAddr));" .

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
@@ -72,23 +72,25 @@ public class AndesKernelBoot {
     /**
      * This will boot up all the components in Andes kernel and bring the server to working state
      */
-    public static void bootAndesKernel() throws AndesException, UnknownHostException, JMException {
+    public static void bootAndesKernel() throws AndesException {
+        try {
+            isKernelShuttingDown = false;
+            //loadConfigurations - done from outside
+            //startAndesStores - done from outside
+            int threadPoolCount = 1;
+            andesRecoveryTaskScheduler = Executors.newScheduledThreadPool(threadPoolCount);
+            startAndesComponents();
+            startHouseKeepingThreads();
+            syncNodeWithClusterState();
+            registerMBeans();
+            startThriftServer();
+            startMessaging();
+            createSuperTenantDLC();
 
-        isKernelShuttingDown = false;
-        //loadConfigurations - done from outside
-        //startAndesStores - done from outside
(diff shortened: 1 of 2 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
```diff
@@ -72,23 +72,25 @@
...
+            startAndesComponents();
+            startHouseKeepingThreads();
+            syncNodeWithClusterState();
+            registerMBeans();
+            startThriftServer();
+            startMessaging();
+            createSuperTenantDLC();
+
+            Andes.getInstance().startSafeZoneAnalysisWorker();
+        } catch (JMException e) {
```
abeykoon: If any exception was thrown here, that means some part of the broker is not working. Is catching JMS exception is enough? If some error happened here, it might be wrong to start the broker without fatal warnings?

modules/andes-core/broker/src/main/java/org/wso2/andes/server/Broker.java
```diff
@@ -92,250 +97,279 @@
...
+                sysInfoMBean.register();
 
-            Set<Integer> sslPorts = new HashSet<Integer>(options.getSSLPorts());
-            if(sslPorts.isEmpty())
-            {
-                parsePortList(sslPorts, serverConfig.getSSLPorts());
-            }
+                Set<Integer> ports = new HashSet<Integer>(options.getPorts());
+                if(ports.isEmpty())
+                {
```
abeykoon: Wrong quotes. Our standard is 
xxx {
   ppp;
}

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
hemikak: @abeykoon , The caught exception is a JMException which is given while registering several MBean services. If any other exception does occur here, wont it get thrown directly?  . Or else how should I handle it?

hemikak:
References for the fix : http://www.avajava.com/tutorials/lessons/how-do-i-use-a-host-name-to-look-up-an-ip-address.html

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
hemikak: The thrown exception will get logged in the management console. Therefore no need to log it here.

PR #126, opened 2015-05-22
Remove exceptions in retain related methods implemented for cassandra datasource

pumudu88:
this PR will remove not implemented exceptions in retain related methods and add suitable empty lists/ maps until retain feature tested in cassandra.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/store/cassandra/CQLBasedMessageStoreImpl.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/store/cassandra/CQLBasedMessageStoreImpl.java
@@ -27,6 +27,7 @@ import java.net.InetSocketAddress;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/store/cassandra/CQLBasedMessageStoreImpl.java
```diff
@@ -882,9 +882,8 @@ public void storeRetainedMessages(List<AndesMessage> retainList) throws AndesExc
     @Override
     public List<String> getAllRetainedTopics() throws AndesException {
         // TODO: implement this method
-        throw new org.apache.commons.lang.NotImplementedException("CQL base message store methods for" +
-                                                                  " retain feature will be implemented " +
-                                                                  "in next iteration");
+        List<String> retainTopicList = Collections.emptyList();
+        return retainTopicList;
```
hemikak: We can directly return a Collections.emptyList() here. I think IntelliJ normally gives a warning here.

PR #129, opened 2015-05-23
andes moves to java 7

indikasampath2000:


modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
```diff
@@ -134,11 +134,13 @@ public static void messageReceived(IncomingMessage incomingMessage, UUID channel
...
         //Following code is only a performance counter
-        Long localCount = receivedMessageCounter.incrementAndGet();
-        if (localCount % 10000 == 0) {
-            long timetook = System.currentTimeMillis() - last10kMessageReceivedTimestamp;
-            log.info("Received " + localCount + ", throughput = " + (10000 * 1000 / timetook) + " msg/sec, " + timetook);
-            last10kMessageReceivedTimestamp = System.currentTimeMillis();
+        if(log.isDebugEnabled()) {
+            Long localCount = receivedMessageCounter.incrementAndGet();
+            if (localCount % 10000 == 0) {
+                long timetook = System.currentTimeMillis() - last10kMessageReceivedTimestamp;
```
ramith: we can do this with matrices later.

PR #135, opened 2015-05-26
method name change getQOSLevel

AnujaLK:


modules/andes-core/broker/src/main/java/org/wso2/andes/mqtt/MQTTopicManager.java
```diff
@@ -208,7 +208,7 @@ public void removeOrDisconnectTopicSubscription(String mqttClientChannelID, Subs
                 //Will remove the subscriber cluster wide
                 try {
                     //Will indicate the disconnection of the topic
-                    if (action == SubscriptionEvent.DISCONNECT) {
+                    if (action == SubscriptionEvent.DISCONNECT && !(subscriber.getQOSLevel() == 0 && !subscriber.isCleanSession())) {
```
ramith: Do we have a Enum for QOS levels?

AnujaLK: This is already fixed from another PR.

sdkottegoda: !(subscriber.getQOSLevel() == 0 && !subscriber.isCleanSession()) could be simplified to (!subscriber.getQOSLevel() == 0 || subscriber.isCleanSession())

AnujaLK: We used this because it was easy to describe the logic according to specification.

PR #136, opened 2015-05-26
Publisher transaction CQL implementation

Asitha:


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/distruptor/inbound/InboundTransactionEvent.java
```diff
@@ -85,6 +112,13 @@ public InboundTransactionEvent(MessageStore.AndesTransaction transaction, Inboun
...
     public void commit() throws AndesException {
+
+        if(currentBatchSize > maxBatchSize) {
+            currentBatchSize = 0;
+            throw new AndesException("Current enqueued batch size exceeds maximum transactional batch size of " +
+            maxBatchSize + " bytes." );
+        }
+
         if (log.isDebugEnabled()) {
             log.debug("Prepare for commit");
```
ramith: shouldn't we be printing a some information to figure out which transaction is going to get committed?

Asitha: You mean some publisher information? In this scope we don't have such information. We can print commit batch size. Would that make sense?

ramith: unsure that would make sense. But consider this: suppose there are 4 transactions that are happening in parallel (for different publishers) and we have enabled debugs (to troubleshoot , or just for kicks!) . in such a situation seeing "Prepare for commit" 4 times would make sense? or would that improve our ability to debug further? so the more general question is how can we keep track of whose doing what transaction? when? how we are going to troubleshoot potential problems (that we may encounter?!). food for thought !

Asitha: Got your point. At the moment we don't have that ability to track publisher information (something like publisher id). What we can do is have the internal Andes channel id printed here. So that we can distinguish the publisers. But still we won't be able to pin point the publishers.

ramith: understood. I leave it to your judgement on having enough information to debug (at least channel ID)

Asitha: I've added the Andes channel id.

PR #147, opened 2015-06-08
[WIP] MQTT TCP back pressure implementation, code refactoring, fixing issues  

pamod:


modules/andes-core/broker/src/main/java/org/dna/mqtt/moquette/messaging/spi/impl/ProtocolProcessor.java
```diff
@@ -594,7 +594,7 @@ void processUnsubscribe(ServerChannel session, String clientID, List<String> top
             //also will unsubscribe from the kernal
             try {
                 AndesMQTTBridge.getBridgeInstance().onSubscriberDisconnection(clientID,
-                        AndesMQTTBridge.SubscriptionEvent.UNSUBSCRIBE);
+                        AndesMQTTBridge.SubscriptionEvent.UN_SUBSCRIBE);
```
AnujaLK: Should "Unsubscribe" be a single word?

pamod: :+1:  , 'unsusbcribe' sounds more readable, it was initially written as 'un-subscribe' due to it being detected as a typo :)

modules/andes-core/broker/src/main/java/org/wso2/andes/mqtt/MQTTMessageContext.java
```diff
@@ -0,0 +1,129 @@
...
+import org.wso2.andes.kernel.distruptor.inbound.PubAckHandler;
+
+import java.nio.ByteBuffer;
+
+import static org.dna.mqtt.wso2.AndesMQTTBridge.QOSLevel;
+
+/**
+ * Builder pattern to contain parameter values defined through the protocol engine
+ */
+public class MQTTMessageContext {
```
ramith: MQTTMessage seems similar to this. could you check?

modules/andes-core/broker/src/main/java/org/dna/mqtt/wso2/MQTTSubscriptionStore.java
```diff
@@ -0,0 +1,86 @@
...
+            Subscribers subscriber = new Subscribers();
+            subscriber.addNewSubscriber(clientID, newSubscription);
+            localSubscriptions.put(topic, subscriber);
+        } else {
+            subscribers.addNewSubscriber(clientID, newSubscription);
+        }
+
+    }
+
+    @Override
```
hemikak: javaDocs missing.

modules/andes-core/broker/src/main/java/org/wso2/andes/mqtt/MQTTLocalSubscription.java
```diff
@@ -33,6 +33,8 @@
 import java.nio.ByteBuffer;
 import java.util.UUID;
 
+import static org.dna.mqtt.wso2.AndesMQTTBridge.*;
```
hemikak: import grouping is there. I think there is a separate settings for static imports in intellij.

modules/andes-core/broker/src/main/java/org/dna/mqtt/wso2/MQTTSubscriptionStore.java
pamod: :+1:  cannot go with  {@inheritDoc} since the methods defined in the parent class do not have java docs and those methods are based on the mqtt protocol library. 

Will go with block comments to provide an insight on the method responsibility.

modules/andes-core/broker/src/main/java/org/wso2/andes/mqtt/MQTTLocalSubscription.java
pamod: Good catch ! :)

modules/andes-core/broker/src/main/java/org/wso2/andes/mqtt/MQTTMessageContext.java
pamod: MQTTMessage extends AndesMessage which includes attributes generic to both protocols, MQTTMessageContext is created at the bridge where all the protocol specific attributes are being built which will be shared across the layers.

PR #149, opened 2015-06-15
MB-1054 issue fixed.

indikasampath2000:
- public jira url : https://wso2.org/jira/browse/MB-1054


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
```diff
@@ -141,14 +151,36 @@ public static void recoverDistributedSlotMap() throws AndesException {
...
+                    }
+                }
+            });
+            futureSlotRecoveryExecutorList.add(submit);
+        }
+        for (Future slotRecoveryExecutor : futureSlotRecoveryExecutorList) {
+            try {
+                slotRecoveryExecutor.get();
+            } catch (InterruptedException e) {
+                log.error("Error occurred in slot recovery.", e);
```
Asitha: I think it would be best to propagate this RuntimeException. This can be done by calling
```
Thread.currentThread().interrupt();
```
Detailed explanation can be found at http://www.ibm.com/developerworks/library/j-jtp05236/#2.1

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
```diff
@@ -170,10 +204,12 @@ private static void initializeSlotMapForQueue(String queueName)
...
         //setting up timer to print restoring message count and database read count
-        numberOfReadsFromDatabase++;
-        restoreSlotMappingCounter = restoreSlotMappingCounter + messageList.size();
+        databaseReadsCounter++;
+        restoreMessagesCounter = restoreMessagesCounter + messageList.size();
+        databaseReadsCounterMap.put(queueName, databaseReadsCounter);
+        restoreMessagesCounterMap.put(queueName, restoreMessagesCounter);
         Timer timer = new Timer();
-        scheduleTimerToPrintCounter(timer);
+        scheduleTimerToPrintCounter(timer, queueName);
```
Asitha: I think it would be great if we can use a `ScheduledExecutorService` instead of a `Timer`. It's the modern API for this type of tasks.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/cassandra/HectorBasedMessageStoreImpl.java
```diff
@@ -470,20 +476,63 @@ public AndesMessageMetadata getMetaData(long messageId) throws AndesException {
...
+                while (lastMsgId <= lastRecoveryMessageId) {
+                    long nextMsgId = lastMsgId + 1;
+                    lastMsgId = nextMsgId + messageIdDifference;
+                    messagesFromQueue = HectorDataAccessHelper
+                            .getMessagesFromQueue(queueName,
+                                    HectorConstants.META_DATA_COLUMN_FAMILY,
+                                    keyspace, nextMsgId, lastMsgId,
+                                    count, true);
+                    if (messagesFromQueue.size() > 0) {
+                        break;
```
Asitha: With having a break here, we may return a message list much greater than the `count` argument requested by the caller of this method or will return much less than the actual messages starting from `firstMsgId`

I think we are breaking the contract between the caller and callee here (violates design by contract concept). This might lead to subtle issues some day in future. Another check or removing extra messages will be a cost, but I guess it'll be for the worthy cause. WDYT?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
hemikak: +1 for this.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/cassandra/ServerStartupRecoveryUtils.java
```diff
@@ -0,0 +1,85 @@
...
+        long firstMsgId;
+        try {
+            String recoveryStartFrom = AndesConfigurationManager.readValue
+                    (AndesConfiguration.RECOVERY_MESSAGES_START_FROM_DATE);
+            SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
+            Date recoveryDate = dateFormat.parse(recoveryStartFrom);
+            firstMsgId = recoveryDate.getTime();
+        } catch (ParseException e) {
+            throw new AndesException("Error while parsing <startRecoveryFrom> property. " +
+                    "Please verify date format.");
```
Asitha: We might need to log the exception.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
```diff
@@ -141,14 +151,36 @@ public static void recoverDistributedSlotMap() throws AndesException {
      */
     private static void recoverMapsForEachQueue() throws AndesException {
         List<AndesQueue> queueList = contextStore.getAllQueuesStored();
-
-        for (AndesQueue queue : queueList) {
+        List<Future> futureSlotRecoveryExecutorList = new ArrayList<Future>();
+        ExecutorService executorService = Executors.newFixedThreadPool(5);
```
hemikak: May I know why use fixed size thread pool of 5 ?

modules/andes-core/broker/src/main/java/org/wso2/andes/store/cassandra/CQLBasedMessageStoreImpl.java
```diff
@@ -654,29 +660,84 @@ public AndesMessageMetadata getMetaData(long messageId) throws AndesException {
                                                                        long firstMsgId,
                                                                        int count) throws AndesException {
 
+        Timer timer = new Timer();
```
Asitha: Can't we use a Scheduled task instead of a `Timer` ?

modules/andes-core/broker/src/main/java/org/wso2/andes/store/cassandra/ServerStartupRecoveryUtils.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Return slot recovery completion message id to stop seeking through cassandra database
+     *
+     * @return id of the message used to stop slot recovery task
+     */
+    public static long getMessageIdToCompleteRecovery() {
+        long REFERENCE_START = 41L * 365L * 24L * 60L * 60L * 1000L;
+        long ts = System.currentTimeMillis();
+        return (ts - REFERENCE_START) * 256 * 1024;
```
hemikak: Could we get an explanation on these numbers?

modules/andes-core/broker/src/main/java/org/wso2/andes/store/cassandra/CQLBasedMessageStoreImpl.java
```diff
@@ -654,29 +660,84 @@ public AndesMessageMetadata getMetaData(long messageId) throws AndesException {
...
+                    resultSet = execute(statement, "retrieving metadata list from " + storageQueueName +
+                            " with starting msg id " + firstMsgId + " ending message id " + lastMsgId);
+                    metadataList =
+                            new ArrayList<AndesMessageMetadata>(resultSet.getAvailableWithoutFetching());
+
+                    for (Row row : resultSet) {
+                        metadataList.add(getMetadataFromRow(row, row.getLong(CQLConstants.MESSAGE_ID)));
+                    }
+                    if (metadataList.size() > 0) {
+                        break;
```
Asitha: I guess we have the same design by contract issue similar to Hector implementation here as well. At the moment, we don't have an issue with it. But if someone use it in future assuming this will return the list less than or equal to count then it would be an issue.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
indikasampath2000: +1. I changed Timer to ScheduledExecutorService.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/cassandra/HectorBasedMessageStoreImpl.java
indikasampath2000: count parameter used in query to limit the result. query may return number of records which may equal or less than to count. While loop going to execute if first query to the Cassandra return 0 metadata which means it's already read tombstone. There after it's continuously running adding up start message id and end message id until it's meet actual records. If query retrieve some actual record, then there should be way to break the while loop and return that result to caller. Query may return number of messages equal or less than to count at this moment. That is what it's checked 'metadataList.size() > 0'. I can change this to 'metadataList.size() <= count' which is give same effect.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/cassandra/ServerStartupRecoveryUtils.java
indikasampath2000: Done.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/cassandra/CQLBasedMessageStoreImpl.java
indikasampath2000: +1. I changed Timer to ScheduledExecutorService.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/cassandra/CQLBasedMessageStoreImpl.java
indikasampath2000: Same explanation given in above HectorBasedMessageStoreImpl.java is apply to here as well.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/cassandra/ServerStartupRecoveryUtils.java
indikasampath2000: +1. I added comment explaining what these number representing.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/cassandra/HectorBasedMessageStoreImpl.java
indikasampath2000: count parameter used in query to limit the result. query may return number of records which may equal or less than to count. While loop going to execute if first query to the Cassandra return 0 metadata which means it's already read tombstone. There after it's continuously running adding up start message id and end message id until it's meet actual records. If query retrieve some actual record, then there should be way to break the while loop and return that result to caller. Query may return number of messages equal or less than to count at this moment. That is what it's checked 'metadataList.size() > 0'.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
indikasampath2000: There could be multiple storage queues worked before entire cluster went down. We need to recover all remaining message of each storage queue at the first node start up in the cluster. So this is done parallely but Cassandra shouldn't be exhausted because there will be multiple database read call at the same time by different threads. 5 is just a optimal number I decided. I will make it to configurable number so user can increase it based on number of storage queues exist to speed up warm start up.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/cassandra/HectorBasedMessageStoreImpl.java
Asitha: Even though we need to check for the scenario with messages returned less than the count in a single query.

indikasampath2000: I made changes to return exactly count until it reach lastRecoveryMessageId

PR #151, opened 2015-06-16
Fix for messages not delivered after a node crash

Asitha:
When a node with publishers crash subscribers on remaining node couldn't get some messages. The messages are in DB, but not delivered to subscribers untill more messages are published to the same queue.

This is the fix for that issue.
Jira: https://wso2.org/jira/browse/MB-1066

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManagerClusterMode.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManagerClusterMode.java
@@ -18,10 +18,10 @@
 
 package org.wso2.andes.kernel.slot;
 
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.hazelcast.core.IMap;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.wso2.andes.kernel.*;
 import org.wso2.andes.server.cluster.coordination.hazelcast.HazelcastAgent;
 import org.wso2.andes.server.cluster.coordination.hazelcast.custom.serializer.wrapper.HashmapStringTreeSetWrapper;
 
@@ -30,7 +30,15 @@ import org.wso2.andes.server.cluster.coordination.hazelcast.custom.serializer.wr
 import org.wso2.andes.server.cluster.coordination.hazelcast.custom.serializer.wrapper.TreeSetSlotWrapper;
 
 
-import java.util.*;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Slot Manager Cluster Mode is responsible of slot allocating, slot creating,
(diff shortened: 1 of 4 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManagerClusterMode.java
```diff
@@ -84,24 +88,42 @@
...
+    private Set<String> queuesToRecover;
 
+    /**
+     * Creates a slot manager for cluster mode slot coordination
+     */
     private SlotManagerClusterMode() {
 
         HazelcastAgent hazelcastAgent = HazelcastAgent.getInstance();
-
+        removedNode = "";
```
Asitha: We need a node id of the removed node. Currently, we don't depend on the member removed method to select the coordinator. Initialising this to an empty string here in case, scheduled task starts before the member removed is called.
Anyway, this will get removed once we move the coordinator selection logic to member removed and added events

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotMessageCounter.java
```diff
@@ -56,6 +56,16 @@
 
     private static final int SLOT_SUBMIT_LOOP_SKIP_COUNT_THRESHOLD = 10;
 
+    /**
+     * Time between successive slot submit scheduled tasks.
+     */
+    public static final int SLOT_SUBMIT_TIMEOUT = 4000;
```
Asitha: Do we need to configure these externally? Anyway, we didn't have this configurable up to now.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManagerClusterMode.java
ramith: :)

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManagerClusterMode.java
```diff
@@ -815,4 +843,53 @@ public void shutDownSlotManager() {
...
+    public void recoverSlots() {
+
+        int threadPoolCount = 1; // Single thread is suffice for this task
+        ThreadFactory namedThreadFactory =
+                new ThreadFactoryBuilder().setNameFormat("RecoverSlotsThreadPool").build();
+        ScheduledExecutorService recoverSlotScheduler =
+                Executors.newScheduledThreadPool(threadPoolCount, namedThreadFactory);
+        queuesToRecover = slotIDMap.keySet();
+
+        recoverSlotScheduler.schedule(
```
ramith: whats the life time of this thread pool? do we need to shut down it later?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotMessageCounter.java
ramith: whats the use case?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotMessageCounter.java
```diff
@@ -102,11 +112,10 @@ public void run() {
                         } catch (ConnectionException e) {
                             log.error("Error while sending slot deletion safe zone update", e);
                         }
-
                     }
                 }
             }
-        }, 4000, 3000);
+        }, SLOT_SUBMIT_TIMEOUT, SLOT_SUBMIT_TASK_EXEC_PERIOD);
```
ramith: why 4000 and 3000 , why can't it be 4500, 3500 ? (basis for the question is are these number chosen based on educated guess or just random?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotMessageCounter.java
Asitha: In a slow message publishing scenario, this is a delay for each message for delivery. If we publish one message per minute then each message will have to wait till this timeout before it goes through the delivery part. A guaranteed delay for a message in slow message publishing scenario.

ramith: shouldn't the above explanation need to end up a the java doc?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotMessageCounter.java
Asitha: We need to ask @hastef88 on this. IMO this depends on the publisher rate as well. in a slow publishing rate this task comes into play. Then this time depends on the maximum message delay we can have and the number of messages for a submit slot event.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManagerClusterMode.java
Asitha: This is a one-shot action that becomes enabled after the given delay. [1]

[1] http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledExecutorService.html#schedule(java.lang.Runnable,%20long,%20java.util.concurrent.TimeUnit)

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotMessageCounter.java
Asitha: Comment updated.

PR #154, opened 2015-06-17
Code review changes and code improvements.

pumudu88:
This PR will include code review changes and code improvements.


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/RetainedContent.java
```diff
@@ -38,6 +38,13 @@
     private Map<Integer, AndesMessagePart> messagePartCache;
     private long messageID;
 
+    /**
+     * Store new retain content based with given params
+     *
+     * @param messageParts
+     * @param contentLength
```
pamod: Could we describe the params ?

PR #155, opened 2015-06-17
Added message store health check to show in UI.

hemikak:


modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/ClusterManager.java
```diff
@@ -347,4 +349,17 @@ public String getCoordinatorNodeAddress() {
...
+    /**
+     * Gets the broker's storage health status.
+     *
+     * @return true if healthy, else false.
+     */
+    public boolean getStoreHealth() {
+        boolean isMessageStoreOperational = AndesContext.getInstance().getMessageStore().isOperational("testString",
+                System.currentTimeMillis());
+        boolean isAndesContextStoreOperational = AndesContext.getInstance().getAndesContextStore().isOperational
+                ("testString", System.currentTimeMillis());
```
ramith: can we use node Id instead of "testString" (even if things work like this)

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
```diff
@@ -323,7 +323,10 @@ public static void startAndesStores() throws Exception {
         messageStore = new FailureObservingMessageStore(createMessageStoreFromConfig(contextStoreInConfig));
         MessagingEngine messagingEngine = MessagingEngine.getInstance();
         messagingEngine.initialise(messageStore, subscriptionStore);
-        
+
+        // setting the message store in the context store
```
Asitha: Shall we begin the sentence with a capital? :)

modules/andes-core/management/common/src/main/java/org/wso2/andes/management/common/mbeans/ClusterManagementInformation.java
```diff
@@ -47,4 +47,11 @@
     @MBeanAttribute(name = "getAllClusterNodeAddresses", description = "Gets the addresses of the members in a cluster")
     List<String> getAllClusterNodeAddresses();
 
+    /**
+     * Gets the broker's storage health status.
+     *
+     * @return true if healthy, else false.
+     */
+    @MBeanAttribute(name = "getStoreHealth", description = "Gets the broker's health status")
```
ramith: should be like 'Gets the message stores health status"

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/ClusterManager.java
hemikak: I have a doubt whether getting "ClusterResourceHolder.getInstance()" would return null in a non-clustered deployment. Therefore I did not used it. 
Should I check that out and use it ?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
hemikak: +1

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/ClusterManager.java
hemikak: It wont return null. Therefore using as suggested.

PR #159, opened 2015-06-20
Content chunking support for Andes core

Asitha:
There is a new set of handlers added to the inbound disruptor ring to resize content chunks if needed. After that MessagePreProcessor will clone messages if necessary.

This will fix MQTT support for large messages as well. 

Jiras: 
https://wso2.org/jira/browse/MB-972
https://wso2.org/jira/browse/MB-1015
https://wso2.org/jira/browse/MB-1039
https://wso2.org/jira/browse/MB-1071

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPUtils.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPUtils.java
@@ -196,6 +196,7 @@ public class AMQPUtils {
         metadata.setSlot(amqMessage.getSlot());
         metadata.setExpirationTime(amqMessage.getExpiration());
         metadata.setArrivalTime(amqMessage.getArrivalTime());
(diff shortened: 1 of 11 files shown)


ramith:
cudos!!!!! hows the performance implications?

Asitha:
I didn't do a performance test with it. Will do one. But if we have the incoming chunk size same as the internal chunk size, chunking logic avoids the content copying. 

Anyway wrote a unit test too. It would be great if we can get that unit testing thing sorted for the build.

Asitha:
Did some test in my local machine for 0.5MB messages with MySQL got similar results in the range of 15 ~ 16 messages per second. 

So did some testing for the content chunking method call with 1 million calls with different message sizes with original chunk size 65534 bytes and output chunk size 65500 bytes

Time taken for a method call in milliseconds is as follows

| 0.5 MB | 1 MB | 1.5 MB | 2 MB 
|-----------|---------|-----------|--------|
| 0.060 | 0.118 | 0.178 | 0.242 |

PR #160, opened 2015-06-20
[MB-1077] fix error when adding  multiple IP addresses for same host name

pumudu88:
This will fix issue in following jira.
https://wso2.org/jira/browse/MB-1077

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/server/Broker.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/server/Broker.java
@@ -211,7 +211,7 @@ public class Broker
 
                         NetworkTransportConfiguration settings =
                                 new ServerNetworkTransportConfiguration(serverConfig, port,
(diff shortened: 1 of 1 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/server/Broker.java
```diff
@@ -189,7 +189,7 @@ private void startAMQPListener(ApplicationRegistry config, BrokerOptions options
                 } else {
                     bindAddressForHostname = InetAddress.getByName(bindAddressFromBrokerOptions);
                 }
-                String hostName = bindAddressForHostname.getCanonicalHostName();
+                String hostName = bindAddressFromBrokerOptions;
```
a5anka: Without changing variable, can't we use the variable "bindAddressFromBrokerOptions" in line 214 and 239. Since hostname is used in line 218 let's keepthis unchanged.

PR #165, opened 2015-06-24
Large content view issue fixed

Asitha:
Queue browser delivery worker content reading logic had a hard-coded constant value for max content chunk size. Updated this to reflect Andes max content chunk size.

Jira: https://wso2.org/jira/browse/MB-1100


ramith:
Looks fine to me.!

PR #174, opened 2015-06-30
Method, deleteMessages modified to only delete message metadata

sdkottegoda:
Adresses the jira https://wso2.org/jira/browse/MB-1132

Since a foreign key relationship was added between the two tables, message metadata and content, it was unnecessary to delete message content explicitly.


ramith:
looks good to me.

PR #180, opened 2015-07-03
Merge issue fixed

AnujaLK:
Removed unnecessary code


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManagerClusterMode.java
```diff
@@ -175,9 +175,6 @@ public Slot getSlot(String queueName, String nodeId) throws AndesException{
             }
         }
 
-        if (andesContextStore == null) {
```
sdkottegoda: flip null check?

AnujaLK: This line was removed.

PR #181, opened 2015-07-04
[WIP] [MB-1144] Added MBeans for search messages in DLC of a specific queue.

hemikak:


hemikak:
Implementation of the MBeans needs to be added.

PR #182, opened 2015-07-05
[MB-734] - Added default mime type for message browsing on UI.

hemikak:
JIRA - https://wso2.org/jira/browse/MB-734


ramith:
looks good to me.

PR #196, opened 2015-07-10
Move tracking MsgData deletion responsibility to SlotDeliveryWorker

a5anka:


a5anka:
Added modifications suggested in the code review.

PR #198, opened 2015-07-13
Add missing error handling task to Exception catch block

a5anka:
The messages in the retain map was not retried in case of an error. Is it the correct behavior?


Asitha:
Nice use of labels. :+1: 
This should be done for all the PR's, by everyone IMO.

PR #206, opened 2015-07-14
Enabled buffer size based flow control

sdkottegoda:
Addresses jira https://wso2.org/jira/browse/MB-1157

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesChannel.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesChannel.java
@@ -171,9 +171,6 @@ public class AndesChannel {
         globalErrorBasedFlowControlEnabled = false;
         unblockLocalChannel();
     }
-
-    
-    
     
     /**
      * Notify local channel to unblock channel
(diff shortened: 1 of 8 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesMessageMetadata.java
```diff
@@ -443,4 +448,22 @@ public void addProperty(String key, Object value) {
     public Object getProperty(String key) {
         return propertyMap.get(key);
     }
+
+    /**
+     * Method to get the byte size of the message associated with this metadata
+     *
+     * @return content size of the message in bytes
+     */
+    public int getMessageContentSize() {
```
Asitha: I guess you can use [AndesMessageMetadata#getMessageContentLength](https://github.com/wso2/andes/blob/master/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesMessageMetadata.java#L395) for this. No need to implement another method.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesChannel.java
```diff
@@ -266,15 +258,16 @@ public void run() {
...
-     * @return total message chunks
+     * @return total message chunk size in Bytes
      */
-    public static int getTotalChunkCount(List<AndesMessage> messages) {
-        int count = 0;
+    public static int getTotalChunkSize(List<AndesMessage> messages) {
+        int size = 0;
         for (AndesMessage message : messages) {
-            count = count + message.getContentChunkList().size();
+            size = size + message.getMetadata().getMessageContentSize();
```
Asitha: use [AndesMessageMetadata#getMessageContentLength](https://github.com/wso2/andes/blob/master/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesMessageMetadata.java#L395) instead.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/FlowControlManager.java
```diff
@@ -105,14 +105,14 @@
     private boolean shutDownTriggered;
 
     public FlowControlManager() {
-        // Read configured limits
-        globalLowLimit = (Integer) AndesConfigurationManager
+        // Read configured limits and multiply by 1048576 to convert MB into Bytes
+        globalLowLimit = 1048576 * (Integer) AndesConfigurationManager
                 .readValue(AndesConfiguration.FLOW_CONTROL_GLOBAL_LOW_LIMIT);
```
abeykoon: Let us use a constant for 1048576

abeykoon: This can be overflowed. Let us document the max possible value in broker.xml config.

ramith:
Am I correct to assume we are not going to merge this one?

sdkottegoda:
yes

sdkottegoda:
We have evaluated this approach vs the existing mechanism of flow controlling

With the new implementation, we face many problems since we cannot not relate between the number of messages and the total space occupied by messages at the configuration time (Messages with different sizes could be published). Following are the problems faced due to this:

    Flow controlling buffer limits are set for sizes in bytes.
        Mina buffer accepts all the messages until we enable flow controlling. (It is enabled when the total size of the messages read by the broker is higher than the specified limit.)
        For the given limit, the number of messages that could be taken in is high for small messages.
        More time is taken to read and process a large number of small messages.
        Therefore, by the time the broker enables flow controlling, the Mina buffer would have overflowed.

    The ring buffer in the disruptor is initialized with a specified buffer size with a given number of elements.
        If the publisher sends messages with smaller sizes, a large number of messages will be taken in before flow controlling is enabled, and the disruptor will overflow.
    The user cannot relate between the number of messages that are published before flow controlling is enabled (By having a count specified as in the existing approach, the number of messages that are taken in is constant whatever the message size is).

Therefore, we decided to leave the configuration as it is.

PR #210, opened 2015-07-16
Fixing message loss during node kill - failover testing

abeykoon:
Fixing RDBMS query to return the slots when a node is killed. Fixing slot return when coordinator itself is killed. Failover happened successfully without message loss. At times several messages came again as acks did not reach sever at the time of killing, which is acceptable.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/ClusterManager.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/ClusterManager.java
@@ -119,9 +119,7 @@ public class ClusterManager {
             clearAllPersistedStatesOfDisappearedNode(deletedNodeId);
 
             //Reassign the slot to free slots pool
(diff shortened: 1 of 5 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/ClusterManager.java
```diff
@@ -119,9 +119,7 @@ public void memberRemoved(Member node) throws AndesException {
             clearAllPersistedStatesOfDisappearedNode(deletedNodeId);
 
             //Reassign the slot to free slots pool
-            if (AndesContext.getInstance().getClusteringAgent().isCoordinator()) {
-                SlotManagerClusterMode.getInstance().reAssignSlotsWhenMemberLeaves(deletedNodeId);
-            }
+            SlotManagerClusterMode.getInstance().reAssignSlotsWhenMemberLeaves(deletedNodeId);
```
sajinidesilva: Shouldn't this be done by the coordinator node?

PR #219, opened 2015-07-18
Initial commit for broker.xml to support ciphertool and to move amqp/mqtt ssl configurations into broker.xml

hastef88:
This is with reference to jiras : 

https://wso2.org/jira/browse/MB-924 
https://wso2.org/jira/browse/MB-1160

At server startup, AndesConfigurationManager will scan for any encrypted properties within broker.xml and add their decrypted values to a static hashmap. Whenever a property is read, this map will be cross referenced to find the decrypted value. This approach is used because cipher tool is dependent on using an AXIOM xml parser. 

A common data structure is used to parse AMQP / MQTT JKS store configurations into the server. Refer JKSStore.java

AMQP SSL configurations are now parsed through AndesConfiguration instead of ServerConfiguration (QPID code.)

This is dependent on : https://github.com/wso2/carbon-business-messaging/pull/100

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/AndesConfigurationManager.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/AndesConfigurationManager.java
@@ -86,6 +86,11 @@ public class AndesConfigurationManager {
     private static final QName SECURE_VAULT_QNAME = new QName("http://org.wso2.securevault/configuration","secretAlias");
 
     /**
+     * Package which contains custom classes that can be parsed by the AndesConfigurationManager.
+     */
+    private static final String CONFIG_MODULE_PACKAGE = "org.wso2.andes.configuration.modules";
+
+    /**
      * Common Error states
      */
     private static final String GENERIC_CONFIGURATION_PARSE_ERROR = "Error occurred when trying to parse " +
@@ -118,7 +123,12 @@ public class AndesConfigurationManager {
      */
     private static CompositeConfiguration compositeConfiguration;
 
-    private static ConcurrentHashMap<String,String> cipherValueMap;
+    /**
+     * This hashmap is used to maintain any properties that were encrypted with ciphertool. key would be the
+     * secretAlias, and the value would be the decrypted value. This will be cross-referenced when reading properties
+     * from broker.xml.
+     */
+    private static ConcurrentHashMap<String, String> cipherValueMap;
 
     /**
      * Decisive configurations coming from carbon.xml that affect the MB configs. e.g port Offset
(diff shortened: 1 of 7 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/AndesConfigurationManager.java
```diff
@@ -60,14 +81,16 @@
...
+     */
+    private static final QName SECURE_VAULT_QNAME = new QName("http://org.wso2.securevault/configuration","secretAlias");
+
+    /**
      * Common Error states
      */
     private static final String GENERIC_CONFIGURATION_PARSE_ERROR = "Error occurred when trying to parse " +
             "configuration value {0}.";
 
-    private static final String NO_CHILD_FOR_INDEX_IN_PROPERTY = "There was no child at the given index {0} for the " +
```
hastef88: Removing unused property. (Was used for the earlier mqtt authentication based on users in broker.xml.)

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/AndesConfigurationManager.java
```diff
@@ -108,46 +133,33 @@
...
             rootConfiguration.setDelimiterParsingDisabled(true);
-            rootConfiguration.setFileName(ROOT_CONFIG_FILE_PATH + ROOT_CONFIG_FILE_NAME);
+            rootConfiguration.setFileName(brokerConfigFilePath);
             rootConfiguration.setExpressionEngine(new XPathExpressionEngine());
             rootConfiguration.load();
             compositeConfiguration.addConfiguration(rootConfiguration);
-            // Load and combine other configurations linked to broker.xml
-            String[] linkedConfigurations = compositeConfiguration.getStringArray("links/link");
-
-            for (String linkedConfigurationPath : linkedConfigurations) {
```
hastef88: Removed this loop since we don't have any linked configurations other than broker.xml. If we plan to refer other configuration files within broker.xml, we can add this later.

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/AndesConfigurationManager.java
```diff
@@ -212,40 +232,6 @@ public static void initialize(int portOffset) throws AndesException {
...
      * <users>
-     * <user userName="testuser1" password="password1" />
-     * <user userName="testuser2" password="password2" />
-     * </users> scenario.
-     *
-     * @param configurationProperty relevant enum value (e.g.- above scenario -> config
-     *                              .enums.AndesConfiguration.TRANSPORTS_MQTT_PASSWORD)
-     * @param index                 index of the child of whom you seek the property (e.g. above scenario -> 1)
-     */
-    public static <T> T readValueOfChildByIndex(AndesConfiguration configurationProperty, int index) {
```
hastef88: Removing unused method.

PR #221, opened 2015-07-20
Added methods to set and get node id from broker.xml

AnujaLK:
JIRA - https://wso2.org/jira/browse/MB-1090


modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/HazelcastClusterAgent.java
```diff
@@ -140,7 +141,7 @@ private String getNodeIdentifier(Member member) {
      * @return id of the node
      */
     public String getIdOfNode(Member node) {
-        return getNodeIdentifier(node);
+        return node.getStringAttribute("nodeId");
```
hastef88: Makes sense to convert "nodeId" to a constant.

PR #222, opened 2015-07-20
Add secure vault support to the andes client

prabathariyaratna:
This addressed the jira https://wso2.org/jira/browse/MB-1161

### DIFF
--- a/modules/andes-core/client/src/main/java/org/wso2/andes/jndi/PropertiesFileInitialContextFactory.java
+++ b/modules/andes-core/client/src/main/java/org/wso2/andes/jndi/PropertiesFileInitialContextFactory.java
@@ -192,11 +192,10 @@ public class PropertiesFileInitialContextFactory implements InitialContextFactor
 
         if (environment != null) {
 
-            Map<String, String> envVariables = (Hashtable<String, String>) environment;
-            Properties properties = convertToProperties(envVariables);
+            Properties properties = convertToProperties(environment);
             SecretResolver secretResolver = SecretResolverFactory.create(properties);
 
-            for (Object key : envVariables.keySet()) {
+            for (Object key : environment.keySet()) {
 
                 if (secretResolver != null && secretResolver.isInitialized()) {
                     if (secretResolver.isTokenProtected(key.toString())) {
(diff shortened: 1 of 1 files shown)


modules/andes-core/client/src/main/java/org/wso2/andes/jndi/PropertiesFileInitialContextFactory.java
```diff
@@ -180,6 +184,41 @@ protected void createConnectionFactories(Map data, Hashtable environment) throws
...
+                        environment.put(key.toString(), secretResolver.resolve(key.toString()));
+                    }
+                }
+
+            }
+
+        }
+    }
+
+    private static Properties convertToProperties(Map map) {
```
ramith: In practice we document all the methods ( in including parameters).
Shall we do that here too?

modules/andes-core/client/src/main/java/org/wso2/andes/jndi/PropertiesFileInitialContextFactory.java
```diff
@@ -180,6 +184,41 @@ protected void createConnectionFactories(Map data, Hashtable environment) throws
         }
     }
 
+    /**
+     * Resolve carbon secure vault encrypted properties.
+     */
+    private static void resolveEncryptedProperties(Hashtable environment) {
```
ramith: shall we document 'environment' ? 
We do this for all methods we write.

modules/andes-core/client/src/main/java/org/wso2/andes/jndi/PropertiesFileInitialContextFactory.java
```diff
@@ -180,6 +184,46 @@ protected void createConnectionFactories(Map data, Hashtable environment) throws
...
+
+    /**
+     * Convert Map to Properties object.
+     * @param map key value pair details
+     */
+    private static Properties convertToProperties(Map map) {
+
+        Properties prop = new Properties();
+
+        for (Object key : map.keySet()) {
```
ramith: question, why not iterate entry list?
and I see usage of this method has the knowledge of map being <String>,<String> and here you are not using that knowledge. any reason?

ramith: my bad. we don't need to iterate the entry list. we only need keys

PR #224, opened 2015-07-20
Fix variable naming issue

a5anka:


Asitha:
Is this why slots were not deleted?

a5anka:
No, I do not think so. This is called only by [MessagingEngine#purgeMessages](https://github.com/wso2/andes/blob/master/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessagingEngine.java#L315).

PR #228, opened 2015-07-21
Slot recovery only done by coordinator on memberleft

a5anka:


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManagerClusterMode.java
```diff
@@ -292,7 +284,7 @@ public void updateMessageID(String queueName, String nodeId, long startMessageId
             firstMessageId = startMessageIdInTheSlot;
         }
 
-	    if (slotRecoveryScheduled.compareAndSet(true, true)) {
```
ramith: why?

a5anka: Using compareAndSet is redundant here since we just want to read (compare) the value here. i.e setting true when true is redundant. isn't it?

ramith: shouldn't the comparison be atomic?

ramith: now your comparison is not atomic isn't it?

a5anka: I think it is. Let me explain,

a. If we use compareAndSet following are the micro steps
 1. compareAndSet will atomically compare and return boolean
 2. if condition will act according to that

b. if we use get,
 1. get will return the boolean value
 2. if condition will act according to that

Isn't it the same?

ramith: does above explanation align with : http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/atomic/AtomicBoolean.html#compareAndSet(boolean,%20boolean)

a5anka: Yes. The important thing to note is we are not setting a new value here. Therefore get() == compareAndSet() (Just for this special case).

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/coordination/hazelcast/HazelcastAgent.java
```diff
@@ -583,14 +583,17 @@ public void setNodeToLastPublishedId(String nodeId, long lastPublishedId) throws
         this.lastPublishedIDMap.set(nodeId, lastPublishedId);
     }
 
+    @Override
```
hemikak: {@inheritDoc} comment.

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/coordination/rdbms/RDBMSAgent.java
```diff
@@ -126,6 +126,11 @@ public void setNodeToLastPublishedId(String nodeId, long lastPublishedId) throws
 		andesContextStore.setNodeToLastPublishedId(nodeId, lastPublishedId);
 	}
 
+	@Override
```
hemikak: {@inheritDoc} comment.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/FailureObservingAndesContextStore.java
```diff
@@ -629,6 +630,16 @@ public void setNodeToLastPublishedId(String nodeId, long messageId) throws Andes
         }
     }
 
+    @Override
```
hemikak: {@inheritDoc} comment.

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/coordination/hazelcast/HazelcastAgent.java
a5anka: Thank you. I will fix those.

PR #230, opened 2015-07-21
Non durable topic delivery path changes

indikasampath2000:
- Non durable topic delivery path changes revert back to original state to avoid message loss by dropping messages to subscriber who doesn't has room to accept message to deliver.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessageFlusher.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessageFlusher.java
@@ -405,30 +405,27 @@ public class MessageFlusher {
                 for (int j = 0; j < subscriptions4Queue.size(); j++) {
                     LocalSubscription localSubscription = findNextSubscriptionToSent(destination,
                             subscriptions4Queue);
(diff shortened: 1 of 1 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessageFlusher.java
```diff
@@ -444,23 +441,24 @@ public int sendMessagesToSubscriptions(String destination, Set<AndesMessageMetad
...
-                    if(numOfCurrentMsgDeliverySchedules == 0) {
-                        AndesRemovableMetadata removableMetadata = new AndesRemovableMetadata(message.getMessageID(),
-                                message.getDestination(), message.getStorageQueueName());
-                        droppedTopicMessagesListRemovable.add(removableMetadata);
-                        droppedTopicMessagesList.add(message);
-                    }
-                    if (log.isDebugEnabled()) {
-                        log.debug("Removing Scheduled to send message from buffer. MsgId= " + message.getMessageID());
+                } else { //normal topic message
+                    if (numOfCurrentMsgDeliverySchedules == subscriptions4Queue.size()) {
```
ramith: @hemikak  would have suggested to swap this!

PR #234, opened 2015-07-23
Add tenant support for MQTT.

akalankapagoda:
Add tenant authentication for MQTT.

MQTT publish or subscribe methods does not allow to provide authentication information. Therefore keeps MQTT client information in memory when connecting and use those information when publishing or subscribing to do tenant validation. 

These details are removed when the client is disconnected.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/dna/mqtt/moquette/messaging/spi/impl/ProtocolProcessor.java
+++ b/modules/andes-core/broker/src/main/java/org/dna/mqtt/moquette/messaging/spi/impl/ProtocolProcessor.java
@@ -26,8 +26,10 @@ import org.wso2.andes.configuration.AndesConfigurationManager;
 import org.wso2.andes.configuration.enums.MQTTUserAuthenticationScheme;
 import org.wso2.andes.kernel.AndesMessageMetadata;
 import org.wso2.andes.kernel.disruptor.inbound.PubAckHandler;
+import org.wso2.andes.mqtt.MQTTAuthorizationSubject;
 import org.wso2.andes.mqtt.MQTTException;
 import org.wso2.andes.mqtt.utils.MQTTUtils;
+import org.wso2.carbon.utils.multitenancy.MultitenantUtils;
 
 import java.nio.ByteBuffer;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


modules/andes-core/broker/src/main/java/org/dna/mqtt/moquette/messaging/spi/impl/ProtocolProcessor.java
```diff
@@ -178,6 +185,9 @@ void processConnect(ServerChannel session, ConnectMessage msg) {
             processPublish(pubEvt);
         }
 
+        MQTTAuthorizationSubject authSubject = new MQTTAuthorizationSubject();
+        authSubject.setClientID(msg.getClientID());
+        authSubject.setUserFlag(msg.isUserFlag());
```
a5anka: Shall we add these two as constructor parameters?

modules/andes-core/broker/src/main/java/org/dna/mqtt/moquette/messaging/spi/impl/ProtocolProcessor.java
```diff
@@ -688,19 +757,21 @@ void processSubscribe(ServerChannel session, SubscribeMessage msg, String client
             }
         }
 
-        //ack the client
-        SubAckMessage ackMessage = new SubAckMessage();
-        ackMessage.setMessageID(msg.getMessageID());
+        // Ack the client only if all subscriptions are successful
+        if (authenticated) {
```
a5anka: When a single subscription in the subscriptions list fails this flag will be false, resulting in not sending a SubAckMessage. Shouldn't we send a SubAckMessage if at least one subscription is successful?

akalankapagoda: If we send success, the subscription which was failed also will think that it was successful. Not sure which is the correct way?

modules/andes-core/broker/src/main/java/org/dna/mqtt/moquette/messaging/spi/impl/ProtocolProcessor.java
akalankapagoda: Sure. Makes more sense that way.

ramith:
@akalankapagoda  thanks for the PR buddy. I merged @a5anka  PR ( which is based on this)

PR #235, opened 2015-07-23
Remove conditional exception throw in catch clause

indikasampath2000:


Asitha:
Have we fixed those errors we get, with Hazelcast based slot management, when shutting down the server.

indikasampath2000:
Hazelcast instance not active exception occurred due to it's shutdown hook trigger when we gracefully shutting down server. We have disable it by adding hazelcast.shutdownhook.enabled=false property to hazelcast.properties. We need to find and fix if we getting same exception further.

PR #247, opened 2015-07-28
Add tenant authentication for MQTT.

a5anka:
MQTT publish or subscribe methods does not allow to provide
authentication information. Therefore keeps MQTT client
information in memory when connecting and use those information
when publishing or subscribing to do tenant validation.

These details are removed when the client is disconnected.

Modified changes done by @akalankapagoda, to send a sub ack if at
least one subscription is allowed for the user.


modules/andes-core/broker/src/main/java/org/dna/mqtt/moquette/messaging/spi/impl/ProtocolProcessor.java
```diff
@@ -283,27 +298,41 @@ protected void processPublish(PublishEvent evt) {
...
-            m_storageService.persistQoS2Message(publishKey, evt);
-            sendPubRec(evt.getClientID(), evt.getMessageID());
-            //We do not add the message to andes at this point since the message addition would happen upon the PUBREL
-        }else{
-            AndesMQTTBridge.onMessagePublished(topic, qos.ordinal(), evt.getMessage(), evt.isRetain(),
-                    evt.getMessageID(), clientID, this,evt.getSession().getSocketChannel());
+            String publishKey;
 
+            // For QOS 2 send publisher received
+            if (qos == AbstractMessage.QOSType.EXACTLY_ONCE) {
```
ramith: Can we flip this?

PR #251, opened 2015-08-03
Added identifer and destination name to flow control message

AnujaLK:
JIRA - https://wso2.org/jira/browse/MB-1122

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/mqtt/connectors/PersistenceStoreConnector.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/mqtt/connectors/PersistenceStoreConnector.java
@@ -94,6 +94,15 @@ public class PersistenceStoreConnector implements MQTTConnector {
                 publisherTopicCorrelate.put(messageContext.getPublisherID(), publisher);
                 //Finally will register the publisher channel for flow controlling
                 AndesChannel publisherChannel = Andes.getInstance().createChannel(publisher);
+                //Set channel details
+                //Substring to remove leading slash character from remote address
+                try {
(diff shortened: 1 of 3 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/mqtt/connectors/PersistenceStoreConnector.java
```diff
@@ -94,6 +94,15 @@ public void addMessage(MQTTMessageContext messageContext) throws MQTTException {
                 publisherTopicCorrelate.put(messageContext.getPublisherID(), publisher);
                 //Finally will register the publisher channel for flow controlling
                 AndesChannel publisherChannel = Andes.getInstance().createChannel(publisher);
+                //Set channel details
+                //Substring to remove leading slash character from remote address
+                try {
+                    publisherChannel.setIdentifier(
+                            messageContext.getChannel().remoteAddress().toString().substring(1));
+                } catch (NullPointerException e) {
+                    publisherChannel.setIdentifier("MQTT-Unknown");
```
ramith: @pamod : any reason why we do this?

pamod: IMO this should perform a null check and should not allow creation of the publisher or perhaps should trigger a warning. @AnujaLK  WDYT ? any idea why we're doing this ?

AnujaLK: Fixed with PR https://github.com/wso2/andes/pull/276

PR #255, opened 2015-08-03
[MB-1174] Prevent node from starting if using 0.0.0.0 as thrift server host

AnujaLK:
JIRA - https://wso2.org/jira/browse/MB-1174

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/MBThriftServer.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/MBThriftServer.java
@@ -65,8 +65,13 @@ public class MBThriftServer {
     public void start(final String hostName,
                       final int port,
                       final String taskName) throws AndesException {
+        /**
+         * Stop node if 0.0.0.0 used as thrift server host in broker.xml
+         */
+        if ("0.0.0.0".equals(hostName)) {
+            throw new RuntimeException("Invalid thrift server host 0.0.0.0");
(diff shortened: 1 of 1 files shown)


pumudu88:
can't we do this validation in AndesConfigurationManager?
since we only validating 0.0.0.0 here, can we set localhost address(127.0.0.1) by deafult if some invalid ip is given? For examle we need to validate for * as well.

AnujaLK:
Validating and throwing exception in MBThriftServer instead of AndesConfigurationManager will avoid adding unwanted tasks to configuration manager.

modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/MBThriftServer.java
```diff
@@ -65,8 +65,13 @@ private MBThriftServer() {
     public void start(final String hostName,
                       final int port,
                       final String taskName) throws AndesException {
+        /**
+         * Stop node if 0.0.0.0 used as thrift server host in broker.xml
+         */
+        if ("0.0.0.0".equals(hostName)) {
+            throw new RuntimeException("Invalid thrift server host 0.0.0.0");
```
ramith: why not throw a andesexception?

AnujaLK: Fixed with PR https://github.com/wso2/andes/pull/276

PR #256, opened 2015-08-04
Fix slot overlapping issue in single node

a5anka:


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotMessageCounter.java
```diff
@@ -196,20 +194,23 @@ private Slot updateQueueToSlotMap(AndesMessageMetadata metadata) {
...
-                slotTimeOutMap.remove(storageQueueName);
-
-            } catch (ConnectionException e) {
-                 /* we only log here since this thread will be run every 3
-                seconds*/
-                log.error("Error occurred while connecting to the thrift coordinator " + e
-                        .getMessage(), e);
+            Long slotStartTime = slotTimeOutMap.get(storageQueueName);
+
+            if (slot.getMessageCount() >= slotWindowSize || System.currentTimeMillis() - slotStartTime >= timeOutForMessagesInQueue) {
```
hemikak: Can we get a comment regarding the if condition ?

a5anka: :+1:  I'll add a comment. I thought it is self explanatory.

PR #258, opened 2015-08-04
fix safe zone calculation logic. 

abeykoon:
safe zone calculation should never stop if the node is live. 
If the network is gone for the node, on exceptions it should backoff, and retry if when the network came back. 

Safe zone calculation is done only by coordinator node. If the coordination was removed from cluster, other nodes will detect member left, a new coordinator will be elected and that coordinator will remove the removed node's entry from safe zone calc. Otherwise safe zone will not proceed and slots will not removed from cluster.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeleteSafeZoneCalc.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeleteSafeZoneCalc.java
@@ -68,8 +68,13 @@ public class SlotDeleteSafeZoneCalc implements Runnable {
                 try {
                     nodesWithPublishedMessages = SlotManagerClusterMode.getInstance().getMessagePublishedNodes();
                 } catch (AndesException e) {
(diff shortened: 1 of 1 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeleteSafeZoneCalc.java
```diff
@@ -68,8 +68,13 @@ public void run() {
...
                     nodesWithPublishedMessages = SlotManagerClusterMode.getInstance().getMessagePublishedNodes();
                 } catch (AndesException e) {
-                    log.error("SlotDeleteSafeZoneCalc stopped due to failing to get message published nodes.", e);
-                    this.setLive(false);
+                    log.error("SlotDeleteSafeZoneCalc stopped due to failing to get message published nodes. "
+                           + "Retrying after 15 seconds" ,e);
+                    try {
+                        Thread.sleep(15 * 1000);
+                    } catch (InterruptedException e1) {
+                        //ignore
```
Asitha: On interrupt should we ignore? [1]

[1] http://www.ibm.com/developerworks/library/j-jtp05236/

PR #269, opened 2015-08-09
[MB-1252] Introduce message delivery model

abeykoon:
Jira: https://wso2.org/jira/browse/MB-1252

1. introduced DeliverableAndesMetadata. This extends AndesMetadata and taking all slot information and delivery information along with it. This prevents lookup of a message bean to get the delivery status. If we do such lookups the map containing the message beans becomes hot in runtime. 

2. Introduced a delivery model for message delivery. A Slot has a life cycle. Likewise a message has a lifecycle (MessageStatus). Inside a message individual delivery channels has their own life cycles (ChannelMessageStatus). They are coupled by SENT_TO_ALL and ACKED_BY_ALL

3. AndesAckData now bares only channel ID ack is received and the DeliverableAndesMessageMatadata reference that we use to deliver the message. 

4. RemovableMessageMatada class is removed. Now two delete messages methods are introduced. One is giving a list of  DeliverableAndesMessageMatadata and other is by passing a AndesMessageMatadata. We pass the references of the messages sent. So the whole lifecycle is maintained in a single object.


modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPUtils.java
```diff
@@ -22,12 +22,7 @@
 import org.wso2.andes.configuration.AndesConfigurationManager;
 import org.wso2.andes.configuration.enums.AndesConfiguration;
 import org.wso2.andes.framing.AMQShortString;
-import org.wso2.andes.kernel.AndesAckData;
-import org.wso2.andes.kernel.AndesContent;
-import org.wso2.andes.kernel.AndesException;
-import org.wso2.andes.kernel.AndesMessageMetadata;
-import org.wso2.andes.kernel.AndesMessagePart;
-import org.wso2.andes.kernel.MessagingEngine;
+import org.wso2.andes.kernel.*;
```
Asitha: Shall we expand this imports?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/Andes.java
```diff
@@ -394,14 +414,28 @@ public void updateMetaDataInformation(String currentQueueName, List<AndesMessage
...
     /**
+     * Schedule message for subscription
+     *
+     * @param messageMetadata message to be scheduled
+     * @param subscription    subscription to send
+     * @throws AndesException
+     */
+    public void reQueueMessageToSubscriber(DeliverableAndesMetadata messageMetadata, LocalSubscription subscription)
+            throws
+            AndesException {
```
Asitha: Shall we move this to a single line?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/DeliverableAndesMetadata.java
```diff
@@ -0,0 +1,544 @@
...
+    private List<MessageStatus> messageStatus;
+    /**
+     * Parent slot of message.
+     */
+    private Slot slot;
+
+    /**
+     * Time stamp message is read from the store
+     */
+    private long timeMessageIsRead;
```
Asitha: Shall we make this field final? We don't want this to change after creation of the object right?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/DeliverableAndesMetadata.java
```diff
@@ -0,0 +1,544 @@
...
+     * @return true if delivered to all the channels
+     */
+    public boolean isMarkAsDelivered() {
+        boolean isDelivered = true;
+        for (Map.Entry<UUID, ChannelInformation> channelInfoEntry : channelDeliveryInfo.entrySet()) {
+            ChannelMessageStatus channelMessageStatus = channelInfoEntry.getValue().getLatestMessageStatus();
+            if(null == channelMessageStatus
+                    || !(channelMessageStatus.equals(ChannelMessageStatus.SENT)
+                    || channelMessageStatus.equals(ChannelMessageStatus.RESENT)
+                    || channelMessageStatus.equals(ChannelMessageStatus.ACKED))) {
```
Asitha: Won't this condition check lead to an NPE? IMO this is ok if there are only &&. Here we have ||

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/DeliverableAndesMetadata.java
```diff
@@ -0,0 +1,544 @@
...
+
+    /**
+     * Check if this message is acknowledged by all the channels it is delivered to
+     * @return true if message is acknowledged by all the channels
+     */
+    public boolean isMarkAsAcked() {
+        boolean isAcked = true;
+        for (Map.Entry<UUID, ChannelInformation> channelInfoEntry : channelDeliveryInfo.entrySet()) {
+            ChannelMessageStatus messageStatus = channelInfoEntry.getValue().getLatestMessageStatus();
+            if(null == messageStatus || !messageStatus.equals(ChannelMessageStatus.ACKED)) {
```
Asitha: Won't this condition check lead to an NPE? IMO this is ok if there are only &&. Here we have ||

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeliveryWorker.java
```diff
@@ -135,7 +130,6 @@ public void run() {
                             Slot currentSlot = requestSlot(storageQueueName);
                             currentSlot.setDestinationOfMessagesInSlot(destinationOfMessagesInQueue);
                             
-                            
                             /**
                              * If the slot is empty
                              */
```
Asitha: Why are't we doing `if (0 == currentSlot.getMessageCount())` here?

PR #283, opened 2015-08-18
Fix for MB-762

a5anka:
Related JIRA https://wso2.org/jira/browse/MB-762

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/DLCQueueUtils.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/DLCQueueUtils.java
@@ -27,10 +27,14 @@ import org.wso2.andes.server.ClusterResourceHolder;
 import org.wso2.andes.server.cluster.coordination.ClusterCoordinationHandler;
 import org.wso2.andes.server.cluster.coordination.hazelcast.HazelcastAgent;
 import org.wso2.andes.kernel.AndesConstants;
+import org.wso2.andes.server.message.ServerMessage;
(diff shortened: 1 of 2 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/DLCQueueUtils.java
```diff
@@ -131,4 +135,31 @@ public static synchronized void createDLCQueue(String tenantName, String tenantO
...
+        ServerMessage amqMessage = message.getMessage();
+        Long messageID = amqMessage.getMessageNumber();
+        String storageQueue = amqMessage.getRoutingKey();
+        AndesRemovableMetadata removableMessage = new AndesRemovableMetadata(messageID, storageQueue, storageQueue);
+
+        List<AndesRemovableMetadata> messageToMoveToDLC = new ArrayList<>();
+        messageToMoveToDLC.add(removableMessage);
+
+        try {
+            log.info("Moving message to Dead Letter Channel. Message ID " + messageID);
```
ramith: should this be a info log? (suppose in a loaded environment suddenly mb put lots of messages (thousands) to DLC and log will have lots of lines.?
(point is if our thing works correctly, this should be a debug log)

ramith: thoughts please.!

a5anka: Yes, I agree. I will make it a debug log.

ramith:
looks ok to me.

PR #288, opened 2015-08-19
[MB-1172] Fix issues in remove non durable subscriptions

pumudu88:
Fix issues occured when removing non-durable subscriptions.

jira - https://wso2.org/jira/browse/MB-1172


ramith:
shall we tag this to a jira?

PR #292, opened 2015-08-21
Change allAcksReceived check

a5anka:
Previouly we use numberOfScheduledDeliveries and
channelToNumOfDeliveries to check if a message can be deleted from
database. In this commit the pendingAckCount is used to track if
all the acks are received. This count is incremented when the
message is given to the outbound disruptor and decremented when an
ack is received.


a5anka:
All the integration tests passed with this change.

PR #293, opened 2015-08-21
cleared messages from  buffer when last local subscriber leaves.

abeykoon:
1. cleared messages from  buffer when last local subscriber leaves.
2. when requeue due to inactive subscriptions, requeue only if there are valid subscriptions for the destination of the message being re-queued


abeykoon:
included in https://github.com/wso2/andes/pull/295

PR #303, opened 2015-08-26
Messages not being moved into DLC issue fixed

sdkottegoda:
Addresses the jira https://wso2.org/jira/browse/MB-1202

The problem was caused by the message delivery count being decremented when a rejection is received. Was resolved by only reducing the 'pendingAckCount' which keeps track on the non-acknowledged message count.


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/OnflightMessageTracker.java
```diff
@@ -203,7 +203,8 @@ public void stampMessageAsRejected(UUID channel, AndesMessageMetadata message) {
         if (trackingData != null) {
```
AnujaLK: Shall we flip the if condition?

sdkottegoda: The class 'OnflightMessageTracker' is removed now.

PR #305, opened 2015-08-27
Fixed NPE when clearing buffered messages

sdkottegoda:
Addresses the issue in https://wso2.org/jira/browse/MB-1205

Added a null check to ensure that an entry with delivery info for each subscription exists before clearing delivery info.

Please merge https://github.com/wso2/product-mb/pull/170 along with this to enable the test cases.


ramith:
what is the root cause for the NPE?
I think we need to look into that as well.

PR #307, opened 2015-08-28
DLC restoring using cached content issue fix

sdkottegoda:
Addresses https://wso2.org/jira/browse/MB-1208

The issue was caused by the usage of the method 'Arrays.asList(array)'
The list returned is not java.util.ArrayList, but a private static class defined inside java.util.Arrays which does not support the remove method.

Please merge https://github.com/wso2/product-mb/pull/174 to enable the test cases

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/server/information/management/QueueManagementInformationMBean.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/server/information/management/QueueManagementInformationMBean.java
@@ -322,7 +322,7 @@ public class QueueManagementInformationMBean extends AMQManagedObject implements
             description = "The Dead Letter Queue Name for the selected tenant") String destinationQueueName) {
 
         if (null != andesMetadataIDs) {
-            List<Long> andesMessageIdList = Arrays.asList(ArrayUtils.toObject(andesMetadataIDs));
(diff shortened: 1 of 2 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/server/information/management/QueueManagementInformationMBean.java
```diff
@@ -322,7 +322,7 @@ public void restoreMessagesFromDeadLetterQueue(@MBeanOperationParameter(name = "
             description = "The Dead Letter Queue Name for the selected tenant") String destinationQueueName) {
 
         if (null != andesMetadataIDs) {
-            List<Long> andesMessageIdList = Arrays.asList(ArrayUtils.toObject(andesMetadataIDs));
+            List<Long> andesMessageIdList = new ArrayList<>(Arrays.asList(ArrayUtils.toObject(andesMetadataIDs)));
```
ramith: Suggestion may be we can do a addAll() -> removes a intermediate asList ?
If this doesn't hit performance-critical path I suggest we modify contract of caching layer.

sdkottegoda: This doesn't hit the performance critical path. This part of the code is only run when a user browses the dead letter channel and restores/re-routes messages from it. Therefore, I did add all. Fixed in the PR https://github.com/wso2/andes/pull/381

PR #309, opened 2015-08-31
Fix NullPointerException in ack handling

a5anka:


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesAckData.java
```diff
@@ -30,6 +30,7 @@
     private String msgStorageDestination;
     private boolean isTopic;
     private UUID channelID;
+    private boolean isRemovable;
```
ramith: shall we document the usage?

ramith: link this PR to a jira.

a5anka: Added doc comments

PR #312, opened 2015-09-01
[MB-1217] Removed messages that are moved to DLC from cache

sdkottegoda:
Addresses the jira https://wso2.org/jira/browse/MB-1217

When messages were moved to DLC, they were removed from the cache.


modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
```diff
@@ -612,6 +612,9 @@ public void moveMetadataToDLC(long messageId, String dlcQueueName) throws AndesE
                 .start();
         Context contextWrite = MetricManager.timer(Level.INFO, MetricsConstants.DB_WRITE).start();
 
+        //Remove the message from cache
+        removeFromCache(messageId);
```
ramith: time taken to remove from cache also going to be on DB_WRITE time.

sdkottegoda: Started  'contextWrite' after removing messages from the cache by the PR https://github.com/wso2/andes/pull/381

PR #316, opened 2015-09-03
[MB-1222] Slot creation timeout fix

AnujaLK:
Jira Id is : https://wso2.org/jira/browse/MB-1222

Initial issue was, messages were not delivered to subscriber when slow publisher disconnected.
Messages were left in database, No slots created. Subscriber did not received messages again until publisher starts publishing messages again.

Cause of this problem was a race condition between concurrently adding and removing timestamp from slotTimeoutMap which is used to create slots using timeout.

We add an entry to slotTimeoutMap only when the new slot is created. But when submitting the previous slot, we clear the slotTimeoutMap. If the clearing happens after timestamp for new slot is added, new entry will be deleted. In that case, it won't be able to submit the new slot.


ramith:
please add description on what the issues is etc etc ( in the jira as well)
helps us to keep track

ramith:
please send a PR to https://github.com/wso2/andes/tree/introduce_new_delivery_model as well.

PR #321, opened 2015-09-07
[MB-1228] Fix intermittent message loss in MQTT subscriber

a5anka:
This issue was observed when the messages were published using two
producers, and consumed using two subscribers (QOS 0, clean
session false)

https://wso2.org/jira/browse/MB-1228


ramith:
looks good to me.

ramith:
what about the small perf related change we discussed about.?

PR #328, opened 2015-09-08
Add stale flag to indicate tracking data deletion

a5anka:


Asitha:
When is a message is marked as stale? What happens to a stale message? Just curious :-)

a5anka:
Stale flag is used to mark a message already in the outbound disruptor when all the subscribers for message destination leaves (i.e. when the slot becomes orphaned). This is to avoid messages being duplicated.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/DeliverableAndesMetadata.java
```diff
@@ -49,6 +49,11 @@
      */
     private long timeMessageIsRead;
 
+    /**
+     * Indicate if the metadata should not be used.
+     */
+    private boolean stale;
```
ramith: shall we explain the usage properly?

a5anka: Done in https://github.com/wso2/andes/pull/335/.

PR #330, opened 2015-09-09
[MB-1246] Fixed durable sub failover when node ID is set in broker.xml

hemikak:
JIRA - https://wso2.org/jira/browse/MB-1246


hemikak:
During DT failover the node ID should be taken from the broker.xml

ramith:
is this only for Durable topics or its any fail over scenario?

hemikak:
It will be modify the memberRemoved function to use the correct node ID. I.E in remove persisted states, slot assignment.

PR #332, opened 2015-09-09
Bug fixes

abeykoon:
fixed subscription close on null pointer if ack receved after closing subscriber. Fixed MQTT bugs.


Asitha:
How can this happen? Even with Disruptor?

PR #339, opened 2015-09-10
Changes related to JIRA MB-1230

ThilankaBowala:
Changes related to https://wso2.org/jira/browse/MB-1230
Related to broker.xml ; in AndesConfiguration and MessagingEngine

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
@@ -286,9 +286,8 @@ public enum AndesConfiguration implements ConfigurationProperty {
             Integer.class),
 
     /**
-     * Published message information is sent to slot coordinator by the node when it either reaches the
-     * slot window size or the window creation timeout. This configures the timeout for slot window creation
-     * task.
+     * Published message information is sent to slot coordinator by the node when it either reaches the slot window
+     * size or the window creation timeout in milliseconds. This configures the timeout for slot window creation task.
      * <p>
      * In a slow message publishing scenario, this is the delay for each message for delivery.
      * For instance if we publish one message per minute then each message will have to wait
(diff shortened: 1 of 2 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessagingEngine.java
```diff
@@ -158,14 +150,6 @@ public void initialise(MessageStore messageStore,
                                           .build();
         asyncStoreTasksScheduler = Executors.newScheduledThreadPool(threadPoolCount , namedThreadFactory);
 
-        // This task will periodically flush message count value to the store
-        messageCountFlusher = new MessageCountFlusher(messageStore, messageCountFlushNumberGap);
-
```
ramith: can't get rid of this class?

ramith: @Asitha  could you have a look.

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
```diff
@@ -287,7 +287,8 @@
 
     /**
      * Published message information is sent to slot coordinator by the node when it either reaches the
-     * slot window size or the window creation timeout. This configures the timeout for slot window creation
+     * slot window size or the window creation timeout in milliseconds. This configures the timeout for slot window
+     * creation
      * task.
```
a5anka: I think we can move this word to the line above.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessagingEngine.java
Asitha: Even though we got rid of further development on Cassandra stores we haven't removed it right? So Cassandra stores depends on this message counter to give message counts for queues. Removing this will result in Cassandra store message counts returning incorrect message counts. That's the impact of removing this class.

PR #341, opened 2015-09-11
[MB-1309] Add trustore configurations when creating the SSLContext in server

a5anka:
https://wso2.org/jira/browse/MB-1309


ramith:
@a5anka shall we attach this to a jira?

a5anka:
Created JIRA MB-1309 to track this

PR #342, opened 2015-09-11
Message duplicate/loss issue fixed in pub/sub close.

indikasampath2000:
- Publisher and subscriber create and close connection for each message and did some test rounds. This was cause to duplicate and loss messages in new delivery model. It has fixed with changes in this commit.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/DeliverableAndesMetadata.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/DeliverableAndesMetadata.java
@@ -368,6 +368,9 @@ public class DeliverableAndesMetadata extends AndesMessageMetadata{
                 break;
             }
         }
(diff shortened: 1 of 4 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/DeliverableAndesMetadata.java
```diff
@@ -368,6 +368,9 @@ public boolean isMarkAsAcked() {
                 break;
             }
         }
+        if(channelDeliveryInfo.entrySet().isEmpty()) {
+            isAcked = false;
```
ramith: channelDeliveryInfo.isEmpty() ?

indikasampath2000: if statement condition changed in https://github.com/wso2/andes/pull/343

PR #344, opened 2015-09-14
[MB-1260] Use ConcurrentHashMaps for slot recovering datastructures

a5anka:
This implements slot recovering task without a scheduled task.

https://wso2.org/jira/browse/MB-1260

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotCreator.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotCreator.java
@@ -1,0 +1,141 @@
+/*
+ * Copyright (c) 2015, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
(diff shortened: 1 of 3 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotCreator.java
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * SlotCreator is used to recover slots belonging to a storage queue when the cluster is restarted.
+ */
+public class SlotCreator implements Runnable {
+
+    /**
+     * Interval between two consecutive stat logs in milliseconds
+     */
+    private static final int STAT_PUBLISHING_INTERVAL = 20 * 1000;
```
a5anka: Can we agree to 20 seconds as the default value?

ramith: 10 seconds?

a5anka: :+1: Default value changed to 10 seconds.

ramith:
looks cool to me.

PR #348, opened 2015-09-15
message duplicate issue fixed in subscriber create and close connection per message

indikasampath2000:
- duplicate messages received when subscriber create and close connection per message. It has fixed with this changes.
- new tracing inforamation added to Message Tracer to identify message requeue to buffer and message discarded when subscriber close
- Message status check changed to avoid sent to all to sent to all and slot return to sent to all


ramith:
cudos ! :)

ramith:
shall we attach a jira to this... this is important piece of work!

indikasampath2000:
Please find the public JIRA related to issue : https://wso2.org/jira/browse/MB-1291

PR #349, opened 2015-09-16
Fix multiple durable topic subscription entries in DB

abeykoon:
fixed issue of adding multiple subscription entries in DB for durable topic subscriptions.


modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/SubscriptionStore.java
```diff
@@ -624,6 +624,50 @@ public synchronized void createDisconnectOrRemoveLocalSubscription(LocalSubscrip
     }
 
     /**
+     * Update local subscription in database
+     *
+     * @param subscription  updated subscription
+     * @throws AndesException
+     */
+    public void updateLocalSubscriptionSubscription(LocalSubscription subscription) throws AndesException {
```
hemikak: Method name is unusual.

PR #352, opened 2015-09-17
updated message delivery model.

abeykoon:
Message status - Removed SENT_TO_ALL status
Channel Message Status  - Removed SENT, RESENT status

Moved purge rule and message expiry rules to the kernel. Now we have two types of delivery rules

AMQPDeliveryRule - AMQP protocol specific
CommonDeliveryRule - kernel level delivery rule

Now state updates are not done inside outbound disruptor. If a send failure happens we detect two types of exceptions and operate on them. 

ProtocolDeliveryFailureException - protocol could not publish - requeue message if durable
ProtocolDeliveryRulesFailureException - protocol level delivery rules failed - send message to DLC if durable


ramith:
This doesn't seem to fix the problem.

PR #357, opened 2015-09-21
[MB-1287] MB hanged when shutting down

a5anka:
MB can hang if the tasks are continuously put in to outbound disruptor when it tries to shutdown. This should not happen normally. But in an error scenario this can make MB hang. Therefore added a timeout to avoid MB being hanged due to this.

https://wso2.org/jira/browse/MB-1287


ramith:
+1 looks good to me.

hemikak:
+1

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/disruptor/delivery/DisruptorBasedFlusher.java
```diff
@@ -140,7 +148,11 @@ public void submit(LocalSubscription subscription, DeliverableAndesMetadata meta
      * before calling this method, otherwise it may never return.
```
Asitha: Shall we update the method comment?

PR #360, opened 2015-09-21
[MB-1314] NullPointerException when subscriber closes connection without consuming messages

a5anka:
Initialize the message flusher when slot delivery worker is given a
destination to deliver messages.

https://wso2.org/jira/browse/MB-1314

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessageFlusher.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessageFlusher.java
@@ -233,6 +233,23 @@ public class MessageFlusher {
         return messageDeliveryInfo;
     }
 
+    /**
+     * Initialize message flusher for delivering messages for the given destination. This is the destination consumers
+     * subscribe.
+     *
+     * @param destination
+     *         Delivery Destination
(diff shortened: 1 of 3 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessageFlusher.java
```diff
@@ -233,6 +233,22 @@ public MessageDeliveryInfo getMessageDeliveryInfo(String destination) throws And
         return messageDeliveryInfo;
     }
 
+    /**
+     * Initialize message flusher for delivering messages for the given destination
+     *
+     * @param destination
```
ramith: is this a storage queue name?

a5anka: No. It is the message destination.

ramith: documentation?

a5anka: Added

PR #364, opened 2015-09-22
[MB-1346] Refactored code in RDBMSConstants

sdkottegoda:
Addresses https://wso2.org/jira/browse/MB-1346

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSConstants.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSConstants.java
@@ -146,316 +146,300 @@ public class RDBMSConstants {
 
     // prepared statements for Message Store
     protected static final String PS_INSERT_MESSAGE_PART =
-            "INSERT INTO " + CONTENT_TABLE + "(" +
-                    MESSAGE_ID + "," +
-                    MSG_OFFSET + "," +
-                    MESSAGE_CONTENT + ") " +
-                    "VALUES (?, ?, ?)";
-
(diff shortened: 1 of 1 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSConstants.java
```diff
@@ -146,316 +146,300 @@
 
     // prepared statements for Message Store
     protected static final String PS_INSERT_MESSAGE_PART =
-            "INSERT INTO " + CONTENT_TABLE + "(" +
-                    MESSAGE_ID + "," +
-                    MSG_OFFSET + "," +
-                    MESSAGE_CONTENT + ") " +
-                    "VALUES (?, ?, ?)";
-
-    protected static final String PS_DELETE_MESSAGE_PARTS =
```
hemikak: This sql is not needed anymore ?

sdkottegoda: No. We don't explicitly delete message parts any more.

a5anka:
:+1:

ramith:
+1

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSConstants.java
ramith: shall we remove.?

sdkottegoda: It is removed by this PR.

PR #370, opened 2015-09-25
Changes related to MB-1276

ThilankaBowala:
Changes related to https://wso2.org/jira/browse/MB-1276

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
@@ -470,9 +470,11 @@ public class AndesKernelBoot {
         log.info("Stop syncing exchanges, queues, bindings and subscriptions...");
         int threadTerminationTimePerod = 20; // seconds
         try {
-            andesRecoveryTaskScheduler.shutdown();
-            andesRecoveryTaskScheduler
-                    .awaitTermination(threadTerminationTimePerod, TimeUnit.SECONDS);
+            if(andesRecoveryTaskScheduler != null){
+                andesRecoveryTaskScheduler.shutdown();
+                andesRecoveryTaskScheduler
+                        .awaitTermination(threadTerminationTimePerod, TimeUnit.SECONDS);
+            }
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
             andesRecoveryTaskScheduler.shutdownNow();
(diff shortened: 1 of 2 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
```diff
@@ -470,9 +470,11 @@ public static void stopHouseKeepingThreads() {
         log.info("Stop syncing exchanges, queues, bindings and subscriptions...");
         int threadTerminationTimePerod = 20; // seconds
         try {
-            andesRecoveryTaskScheduler.shutdown();
-            andesRecoveryTaskScheduler
-                    .awaitTermination(threadTerminationTimePerod, TimeUnit.SECONDS);
+            if(andesRecoveryTaskScheduler != null){
```
a5anka: Do we need this null check?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/disruptor/inbound/InboundKernelOpsEvent.java
```diff
@@ -127,7 +127,9 @@ public void startMessageDelivery() {
      * Stop message delivery threads in Andes
      */
     public void stopMessageDelivery() {
-        messagingEngine.stopMessageDelivery();
+        if(messagingEngine != null){
```
a5anka: Can messagingEngine be null at this point? if it is null why are we calling "stopMessageDelivery"?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
ThilankaBowala: At this point, these are not initialized. So, it gives the NullPointerException. Since it is OK to give that exception when shutting down, I'll remove these null checks.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/disruptor/inbound/InboundKernelOpsEvent.java
ThilankaBowala: Since we are trying to shut down the carbon kernel, messagingEngine is not initialized. But, shutdown method is a general one for all scenarios.

PR #371, opened 2015-09-28
Refactor slot timeout task

a5anka:
- Replaces the Timer used in SlotMessageCounter with a ScheduledExecutorService
- Remove nodeInformedSlotDeletionSafeZones map from Slot Coordinator

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotMessageCounter.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotMessageCounter.java
@@ -18,6 +18,7 @@
 
 package org.wso2.andes.kernel.slot;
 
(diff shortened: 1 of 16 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotMessageCounter.java
```diff
@@ -140,12 +124,11 @@ public void recordMetadataCountInSlot(Collection<AndesMessage> messageList) {
      *
      * @param metadata AndesMessageMetadata
      */
-    public void recordMetadataCountInSlot(AndesMessageMetadata metadata) {
+    private void recordMetadataCountInSlot(AndesMessageMetadata metadata) {
```
ramith: shoudn't we rename this to something like recordMessageCountInSlot or something. given that metadata : message is 1:1 relationship

PR #372, opened 2015-09-29
[MB-1342] Unique constraint violation fix

pumudu88:
This PR will fix following jira.
https://wso2.org/jira/browse/MB-1342

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
@@ -30,6 +30,7 @@ import org.wso2.andes.kernel.DurableStoreConnection;
 import org.wso2.andes.kernel.MessageStore;
 import org.wso2.andes.kernel.slot.Slot;
 import org.wso2.andes.metrics.MetricsConstants;
+import org.wso2.andes.store.AndesDataIntegrityViolationException;
 import org.wso2.andes.store.cache.AndesMessageCache;
 import org.wso2.andes.store.cache.MessageCacheFactory;
 import org.wso2.andes.tools.utils.MessageTracer;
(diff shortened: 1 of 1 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
```diff
@@ -1392,9 +1394,18 @@ private void createNewQueue(final Connection connection, final String destinatio
...
         } catch (SQLException e) {
-            log.error("Error occurred while inserting destination queue [" + destinationQueueName +
-                    "] to database ");
-            throw e;
+            AndesException andesException =
+                    rdbmsStoreUtils.convertSQLException("Error occurred while creating queue", e);
+
+            if(andesException instanceof AndesDataIntegrityViolationException) {
+                // This exception occurred because some other node has created the queue in parallel.
+                // Therefore no need to create the queue. It's already created.
```
ramith: may be we can print a WARN here?

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
```diff
@@ -1392,9 +1394,18 @@ private void createNewQueue(final Connection connection, final String destinatio
...
+                // Nothing need to be done if this exception occur.
+            } else {
+                log.error("Error occurred while inserting destination queue [" + destinationQueueName +
+                          "] to database ");
```
ramith: should we be logging with the exception?

ramith:
can we merge two commits.?

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
pumudu88: +1

PR #373, opened 2015-09-29
Fix Overlapped Slot Assignment.

abeykoon:
Fixing https://wso2.org/jira/browse/MB-1362

Issue here is same slot is assigned to both nodes and duplicates are delivered.


abeykoon:
Outdated. Hence closing.

PR #374, opened 2015-09-30
Fixed MB-1385

sajinidesilva:
https://wso2.org/jira/browse/MB-1385


modules/andes-core/client/src/main/java/org/wso2/andes/client/AMQSession.java
```diff
@@ -3358,7 +3358,7 @@ public void run()
             try
             {
                 Dispatchable disp;
-                while (!_closed.get() && ((disp = (Dispatchable) _queue.take()) != null))
+                while (((disp = (Dispatchable) _queue.take()) != null) && !_closed.get())
```
ramith: could we have more paranthesis to make it more readable?
(!_closed.get())

woudl that make it more unreadable ? :)

PR #379, opened 2015-10-01
Implement getting overlapped slots for a node id

a5anka:


a5anka:
This change is already merged in a different pull request

PR #382, opened 2015-10-01
[MB-1338][MB-1339] Fix port enable/disable issue and refactor code

pumudu88:
Fix for secure/non-secure port enable/disable from broker configurations. Some code refactoring also included.
https://wso2.org/jira/browse/MB-1338
https://wso2.org/jira/browse/MB-1339

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/qpid/ServerConfiguration.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/qpid/ServerConfiguration.java
@@ -567,6 +567,11 @@ public class ServerConfiguration extends ConfigurationPlugin implements SignalHa
                 !(Boolean)AndesConfigurationManager.readValue(AndesConfiguration.TRANSPORTS_AMQP_DEFAULT_CONNECTION_ENABLED);
     }
 
+    public boolean getMQTTSSLOnly() {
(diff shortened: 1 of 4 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/qpid/ServerConfiguration.java
```diff
@@ -567,6 +567,11 @@ public boolean getSSLOnly() {
                 !(Boolean)AndesConfigurationManager.readValue(AndesConfiguration.TRANSPORTS_AMQP_DEFAULT_CONNECTION_ENABLED);
     }
 
+    public boolean getMQTTSSLOnly() {
+        return (Boolean) AndesConfigurationManager.readValue(AndesConfiguration.TRANSPORTS_MQTT_SSL_CONNECTION_ENABLED) &&
+               !(Boolean)AndesConfigurationManager.readValue(AndesConfiguration.TRANSPORTS_MQTT_DEFAULT_CONNECTION_ENABLED);
```
sajinidesilva: Can't we simplify this further so that anyone can understand?

pumudu88: Currently this is how we check if only SSL port enabled in AMQP implementation. https://github.com/pumudu88/andes/blob/fix_mqtt_issues/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/qpid/ServerConfiguration.java#L565-L568
Will refactor this in more suitable way for both AMQP and MQTT in another iteration.

PR #386, opened 2015-10-02
[MB-1268] Implemented message count retrieval for all queues

sdkottegoda:
Addresses the issue reported at https://wso2.org/jira/browse/MB-1268

The issue was caused by a large number of queries being executed on the database. The Method to retrieve all queues was implemented as follows:
1. Get all the queue names from the AMQPConstructStore
2. For each name, get the number of messages in the database from the metadata table

Therefore, a database call was made for each query in the construct store. The solution to the issue was to get the message count in the metadata table grouped by each queue_id by a single db call. This way, the operation to retrieve all queues is converted to a single operation rather than a number of operations retrieving each queue.

Please merge https://github.com/wso2/carbon-business-messaging/pull/171 and https://github.com/wso2/product-mb/pull/213 along with this

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/DLCQueueUtils.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/DLCQueueUtils.java
@@ -72,6 +72,27 @@ public class DLCQueueUtils {
     }
 
     /**
+     * Derive the Dead Letter Queue name of the tenant with respect to a given queue of the same
+     * tenant.
+     * <p/>
+     * The dead letter of the super domain is named as 'DeadLetterChannel' and the DLC of a tenant takes the
+     * format 'tenantDomain/DeadLetterChannel'
+     * Therefore, The DLC queue name return by this method takes either of the above formats.
+     *
+     * @param tenantDomain the domain name of the tenant
+     * @return The Dead Letter Queue name for the tenant.
+     */
+    public static String generateDLCQueueNameFromTenant(String tenantDomain) {
+        //If the tenantDomain is either null of if it is "carbon.super", the DLC of the super domain will be returned
+        //Else, the dlc for the respective tenant will be returned
+        if (null != tenantDomain && !("carbon.super".equals(tenantDomain))) {
+            return tenantDomain + AndesConstants.TENANT_SEPARATOR + AndesConstants.DEAD_LETTER_QUEUE_SUFFIX;
+        } else {
+            return AndesConstants.DEAD_LETTER_QUEUE_SUFFIX;
+        }
+    }
+
+    /**
      * Decides on whether a given queue name is a Dead Letter Queue or not.
      *
      * @param queueName The Queue name to test.
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/store/FailureObservingMessageStore.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/store/FailureObservingMessageStore.java
@@ -437,6 +437,19 @@ public class FailureObservingMessageStore implements MessageStore {
      * {@inheritDoc}
      */
     @Override
+    public Map<String, Integer> getMessageCountForAllQueues(List<String> queueNames) throws AndesException {
+        try {
+            return wrappedInstance.getMessageCountForAllQueues(queueNames);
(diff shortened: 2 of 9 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/DLCQueueUtils.java
```diff
@@ -72,6 +72,27 @@ public static String identifyTenantInformationAndGenerateDLCString(String queueN
...
+     * format 'tenantDomain/DeadLetterChannel'
+     * Therefore, The DLC queue name return by this method takes either of the above formats.
+     *
+     * @param tenantDomain the domain name of the tenant
+     * @return The Dead Letter Queue name for the tenant.
+     */
+    public static String generateDLCQueueNameFromTenant(String tenantDomain) {
+        //If the tenantDomain is either null of if it is "carbon.super", the DLC of the super domain will be returned
+        //Else, the dlc for the respective tenant will be returned
+        if (null != tenantDomain && !("carbon.super".equals(tenantDomain))) {
```
ramith: carbon.super is already define in Carbon level constants. we can use that.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSConstants.java
```diff
@@ -186,6 +186,21 @@
...
 
+    protected static final String ALIAS_FOR_QUEUES = "QUEUE_COUNT";
+
+    /**
+     * Prepared statement to select all the queue names with the number of messages remaining in the database
+     * by joining the tables MB_QUEUE_MAPPING and MB_METADATA.
+     */
+    protected static final String PS_SELECT_ALL_QUEUE_MESSAGE_COUNT =
+            "SELECT " + QUEUE_NAME + ", " + PS_ALIAS_FOR_COUNT
+            + " FROM " + QUEUES_TABLE + " LEFT OUTER JOIN "
```
ramith: have we verified this in all dbs we support?

modules/andes-core/broker/src/main/java/org/wso2/andes/store/FailureObservingMessageStore.java
```diff
@@ -437,6 +437,19 @@ public void addQueue(String storageQueueName) throws AndesException {
      * {@inheritDoc}
      */
     @Override
+    public Map<String, Integer> getMessageCountForAllQueues(List<String> queueNames) throws AndesException {
+        try {
+            return wrappedInstance.getMessageCountForAllQueues(queueNames);
+        } catch (AndesStoreUnavailableException exception) {
+            notifyFailures(exception);
```
akalankapagoda: Can we introduce another method like notifyFailuresAndThrow so we don't have to have this sequence of statements redundantly?

sdkottegoda: Yeah we can. But if we create a notifyFailuresAndThrow with the return type as void, we'll again need to explicitly return an object after calling it. For example, the getMessageCountForAllQueues method will look as follows:

public Map<String, Integer> getMessageCountForAllQueues(List<String> queueNames) throws AndesException {
        try {
            return wrappedInstance.getMessageCountForAllQueues(queueNames);
        } catch (AndesStoreUnavailableException exception) {
            notifyFailuresAndThrow(exception);
            return null;
        }
    }

This adds an unnecessary overhead.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSConstants.java
sdkottegoda: Yeah, it is supported in mysql, mssql and oracle. When I went through reviews for H2, some say it isn't working properly. But I did test it with H2 and it had no issue.

modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/DLCQueueUtils.java
sdkottegoda: Addressed by the PR https://github.com/wso2/andes/pull/403

PR #398, opened 2015-10-06
[MB-1419] Use limited max unacked messages only for topic and restrict client session to have only one consumer

akalankapagoda:
Fixing - https://wso2.org/jira/browse/MB-1419

Use the LimitedSizeQueueEntryHolder only for topic messages and restrict andes client to allow only one consumer per session.
This is because although QPID is designed so that there can be many consumers per session, Andes is designed so that session and consumer is a one-to-one mapping.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/server/AMQChannel.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/server/AMQChannel.java
@@ -142,8 +142,10 @@ public class AMQChannel implements SessionConfig, AMQSessionModel
 
     private final MessageStore _messageStore;
 
+    // Default map to store unacked messages. This will replaced according to the delivery strategy when a subscriber
+    // is registered to this channel
     private UnacknowledgedMessageMap _unacknowledgedMessageMap = new UnacknowledgedMessageMapImpl(DEFAULT_PREFETCH,
-            this);
+            this, false);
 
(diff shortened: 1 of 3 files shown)


modules/andes-core/client/src/main/java/org/wso2/andes/client/AMQSession.java
```diff
@@ -2040,6 +2040,10 @@ protected C createConsumerImpl(final Destination destination, final int prefetch
                                                  final int prefetchLow, final boolean noLocal, final boolean exclusive, String selector, final FieldTable rawSelector,
                                                  final boolean noConsume, final boolean autoClose) throws JMSException
     {
+        if (this._consumers.values().size() > 0 ) {
+            throw new JMSException("Only one consumer can be created per session");
```
a5anka: Should we do this in the client code? Since the client should behave according the JMS spec and since we cannot change other client implementations (RabittMQ etc.) I think we should not do this? May be we can throw and error in the back-end or actually fix it (I don't know how much time it will take :smile:).

WDYT?

akalankapagoda: We cannot support having multiple consumers in a single session with the current architecture. Will try to update the server code.

a5anka: :+1:

PR #410, opened 2015-10-12
[MB-1245] DurableTopicsSubscribers are not getting unsubscribed

sdkottegoda:
Addresses the issue reported at https://wso2.org/jira/browse/MB-1245.

If the user becomes unable to unsubscribe once, he will never be able to unsubscribe since the queue is deleted prior to unsubscribing (the existence of the queue is checked at an early stage of unsubscribing). This PR does not solve the issue completely, but minimizes the issue to some extent by deleting the queue only after unsubscribing, therefore becoming able to attempt once again.


a5anka:
:+1:

ramith:
+1

PR #411, opened 2015-10-12
Fixing MB-1422, The retained flag is not set to true when client recieves a retained message

ramith:
Jira is : https://wso2.org/jira/browse/MB-1422
The retained flag is not set to true when client recieves a retained message


modules/andes-core/broker/src/main/java/org/dna/mqtt/moquette/messaging/spi/impl/ProtocolProcessor.java
```diff
@@ -409,8 +409,8 @@ public void publishToSubscriber(String topic, AbstractMessage.QOSType qos, ByteB
             if (qos == AbstractMessage.QOSType.MOST_ONE && subscription.isActive()) {
                 //QoS 0
                 //We keep the retain flag false always
```
sdkottegoda: I think we need to remove this comment

ramith:
@sdkottegoda done.

PR #412, opened 2015-10-13
Fix mqtt clean session = false delivery path and resolve delivery when a subscriber changes the subscribed topic.

akalankapagoda:
This pull fixes the following issues.

https://wso2.org/jira/browse/MB-1472
https://wso2.org/jira/browse/MB-1471
https://wso2.org/jira/browse/MB-1468
https://wso2.org/jira/browse/MB-1425
https://wso2.org/jira/browse/MB-1334
https://wso2.org/jira/browse/MB-1423

MQTT subscription handling is changed so that when a client is connected to two different topics, Andes will create two subscriptions, one per each topic, manipulating the subscription ID.

Changed the storage queue name for clean session = false subscribers to a combination of clienId + topic, instead of just client Id.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/LocalSubscription.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/LocalSubscription.java
@@ -90,11 +90,16 @@ public class LocalSubscription  extends BasicSubscription implements InboundSubs
 
         this.subscription = subscription;
 
-        //We need to keep the subscriptionType in basic subscription for notification
-        if(subscription instanceof AMQPLocalSubscription) {
-            setSubscriptionType(SubscriptionType.AMQP);
-        } else if(subscription instanceof MQTTLocalSubscription) {
-            setSubscriptionType(SubscriptionType.MQTT);
+        // In case of mock subscriptions, this becomes null
+        if (null != subscription) {
+            //We need to keep the subscriptionType in basic subscription for notification
+            if(subscription instanceof AMQPLocalSubscription) {
+                setSubscriptionType(SubscriptionType.AMQP);
+            } else if(subscription instanceof MQTTLocalSubscription) {
+                setSubscriptionType(SubscriptionType.MQTT);
+            }
+
+            setStorageQueueName(subscription.getStorageQueueName(destination, subscribedNode));
         }
 
         this.maxNumberOfUnAcknowledgedMessages = AndesConfigurationManager.readValue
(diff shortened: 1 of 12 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/OutboundSubscription.java
```diff
@@ -18,9 +18,7 @@
 
 package org.wso2.andes.subscription;
 
-import org.wso2.andes.kernel.AndesContent;
-import org.wso2.andes.kernel.AndesException;
-import org.wso2.andes.kernel.ProtocolMessage;
+import org.wso2.andes.kernel.*;
```
hemikak: Have to expand this.

modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/LocalSubscription.java
```diff
@@ -90,11 +90,16 @@ public LocalSubscription(OutboundSubscription subscription, String subscriptionI
...
-            setSubscriptionType(SubscriptionType.AMQP);
-        } else if(subscription instanceof MQTTLocalSubscription) {
-            setSubscriptionType(SubscriptionType.MQTT);
+        // In case of mock subscriptions, this becomes null
+        if (null != subscription) {
+            //We need to keep the subscriptionType in basic subscription for notification
+            if(subscription instanceof AMQPLocalSubscription) {
+                setSubscriptionType(SubscriptionType.AMQP);
+            } else if(subscription instanceof MQTTLocalSubscription) {
+                setSubscriptionType(SubscriptionType.MQTT);
```
ramith: is there a way to separate out MQTT related stuff from kernel ( may be we can do it in next release)

modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/OutboundSubscription.java
akalankapagoda: yes, will expand.

modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/LocalSubscription.java
akalankapagoda: This is used when resolving wild card subscriptions which are different in MQTT and AMQP. Since both MQTT and AMQP subscriptions have to extend BasicSubscription which is not an interface, we need to change the behaviour of the AndesSubscription class hierarchy. Therefore, It is better to invest some time on this.

PR #413, opened 2015-10-13
Properly propergate queue creation and binding creation errors to andes client

abeykoon:
1. Previously, always "access refused" error was propagated. 
2. Error when creating queues were not propagated.


akalankapagoda:
This pull fixes the following issue.

https://wso2.org/jira/browse/MB-1451

PR #419, opened 2015-10-14
Fixed MB-1466

sajinidesilva:
Tracked by https://wso2.org/jira/browse/MB-1466


modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/ClusterSubscriptionBitMapHandler.java
```diff
@@ -158,6 +152,22 @@ public void addWildCardSubscription(AndesSubscription subscription) throws Andes
...
+     * Replace the old wildcard subscription with the new wildcard subscription
+     * @param subscription The subscription to be updated
+     * @throws AndesException
+     */
+    @Override
+    public void updateWildCardSubscription(AndesSubscription subscription) throws AndesException {
+        for(AndesSubscription existingSubscription : wildCardSubscriptionList){
+            if(subscription.getTargetQueue().equals(existingSubscription.getTargetQueue())){
+                wildCardSubscriptionList.remove(existingSubscription);
+                wildCardSubscriptionList.add(subscription);
```
hemikak: Do we need to 'break" here ? depends on the logic.

PR #423, opened 2015-10-15
[MB-1245] Fixed issue in unsubscribing durable subscriptions

sdkottegoda:
Addresses the issue reported at https://wso2.org/jira/browse/MB-1245.

The issue occurred when trying to unsubscribe durable subscriptions after restarting the whole cluster. The root cause for the issue was that the subscriptions were identified from local subscription maps and the local subscription maps were empty after restarting nodes.

The issue was solved by deleting all cluster subscriptions from the cluster subscription masp regardless of its presence in the local subscription map.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesContextInformationManager.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesContextInformationManager.java
@@ -66,6 +66,11 @@ public class AndesContextInformationManager {
     private SubscriptionStore subscriptionStore;
 
     /**
+     * Manages all operations related to subscription changes such as addition, disconnection and deletion
+     */
+    AndesSubscriptionManager subscriptionManager;
+
+    /**
      * Initializes the andes context information manager
      *
      * @param subscriptionStore The subscriptions store
@@ -75,6 +80,7 @@ public class AndesContextInformationManager {
                                           AndesContextStore contextStore,
                                           MessageStore messageStore) {
 
+        this.subscriptionManager = ClusterResourceHolder.getInstance().getSubscriptionManager();
         this.subscriptionStore = subscriptionStore;
         this.messageStore = messageStore;
         this.contextStore = contextStore;
@@ -187,9 +193,9 @@ public class AndesContextInformationManager {
         //purge the queue cluster-wide
         MessagingEngine.getInstance().purgeMessages(queueName, null, false);
 
-        //delete all subscription entries if remaining (inactive entries)
-        ClusterResourceHolder.getInstance().getSubscriptionManager()
-                .deleteAllLocalSubscriptionsOfBoundQueue(queueName);
+        //delete all local and cluster subscription entries if remaining (inactive entries)
+        subscriptionManager.deleteAllLocalSubscriptionsOfBoundQueue(queueName);
+        subscriptionManager.deleteAllClusterSubscriptionsOfBoundQueue(queueName);
 
         // delete queue from construct store
         constructStore.removeQueue(queueName, true);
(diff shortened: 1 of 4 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesContextInformationManager.java
```diff
@@ -188,8 +188,11 @@ public void deleteQueue(String queueName) throws AndesException {
         MessagingEngine.getInstance().purgeMessages(queueName, null, false);
 
         //delete all subscription entries if remaining (inactive entries)
+        ClusterResourceHolder.getInstance().getSubscriptionManager().deleteAllLocalSubscriptionsOfBoundQueue(queueName);
+
```
a5anka: Better to store "ClusterResourceHolder.getInstance().getSubscriptionManager()" in a field variable since we are reusing it.

modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/SubscriptionStore.java
```diff
@@ -447,6 +447,34 @@ public boolean isSubscriptionAvailable(AndesSubscription subscription) throws An
...
+     * @return List if matching subscriptions
+     * @throws AndesException
+     */
+    public Set<AndesSubscription> getListOfClusterSubscriptionsBoundToQueue(String queueName) throws AndesException {
+        Set<AndesSubscription> subscriptionsOfQueue = new HashSet<>();
+        Set<AndesSubscription> queueSubscriptionMap = clusterQueueSubscriptionMap.get(queueName);
+        if (queueSubscriptionMap != null) {
+            subscriptionsOfQueue.addAll(queueSubscriptionMap);
+        }
+        Map<String, Set<AndesSubscription>> topicSubscriptionMap = clusterTopicSubscriptionMap;
```
a5anka: Do we need to declare a local variable here. Can't we use the clusterTopicSubscriptionMap?

modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/SubscriptionStore.java
```diff
@@ -758,9 +786,11 @@ public LocalSubscription removeLocalSubscription(LocalSubscription subscription)
                 destination = subscriptionToRemove.getSubscribedDestination();
             }
             String destinationIdentifier = (subscriptionToRemove.isBoundToTopic() ? TOPIC_PREFIX : QUEUE_PREFIX) + destination;
-            andesContextStore.removeDurableSubscription(destinationIdentifier, subscription.getSubscribedNode() + "_" + subscriptionID);
+            andesContextStore.removeDurableSubscription(destinationIdentifier,
+                    subscription.getSubscribedNode() + "_" + subscriptionID);
             if (log.isDebugEnabled()) {
-                log.debug("Subscription Removed Locally for  " + destination + "@" + subscriptionID + " " + subscriptionToRemove);
+                log.debug("Subscription Removed Locally for  " + destination + "@" + subscriptionID + " " +
+                          subscriptionToRemove);
```
a5anka: '+' operator should precede. :smile:

modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/SubscriptionStore.java
```diff
@@ -769,6 +799,36 @@ public LocalSubscription removeLocalSubscription(LocalSubscription subscription)
...
+     * Remove cluster subscriptions from database
+     *
+     * @param subscriptionToRemove The the set of andes subscriptions to be removed
+     */
+    public void removeClusterSubscriptions(Set<AndesSubscription> subscriptionToRemove) throws AndesException {
+        Iterator<AndesSubscription> iterator = subscriptionToRemove.iterator();
+        while (iterator.hasNext()) {
+            AndesSubscription subscription = iterator.next();
+            String destination = getDestination(subscription);
+            if (null != subscriptionToRemove) {
```
a5anka: If the subscriptionToRemove is null there will be a NullPointerException at line 807. Therefore no use in checking here.

PR #426, opened 2015-10-16
MB-Fixed 1449 Exception occurred when one subscriber is stopped

sajinidesilva:


sajinidesilva:
Tracked  by https://wso2.org/jira/browse/MB-1449

a5anka:
:+1:

PR #429, opened 2015-10-18
Change MQTT storage queue name and restrict will message

akalankapagoda:
Fix for https://wso2.org/jira/browse/MB-1515.

Change MQTT storage queue name to behave the same as AMQP when it is not durable.
Remove processing of last will messages since it is not yet supported.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/dna/mqtt/moquette/messaging/spi/impl/ProtocolProcessor.java
+++ b/modules/andes-core/broker/src/main/java/org/dna/mqtt/moquette/messaging/spi/impl/ProtocolProcessor.java
@@ -205,12 +205,13 @@ public class ProtocolProcessor implements EventHandler<ValueEvent>, PubAckHandle
 
         //Handle will flag
         if (msg.isWillFlag()) {
-            AbstractMessage.QOSType willQos = AbstractMessage.QOSType.values()[msg.getWillQos()];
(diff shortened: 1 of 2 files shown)


modules/andes-core/broker/src/main/java/org/dna/mqtt/moquette/messaging/spi/impl/ProtocolProcessor.java
```diff
@@ -205,12 +205,13 @@ void processConnect(ServerChannel session, ConnectMessage msg) {
...
-            PublishEvent pubEvt = new PublishEvent(msg.getWillTopic(), willQos,
-                    bb, msg.isWillRetain(), msg.getClientID(), session);
-            processPublish(pubEvt);
+            log.warn("Andes does not support last will operation");
+//            AbstractMessage.QOSType willQos = AbstractMessage.QOSType.values()[msg.getWillQos()];
+//            byte[] willPayload = msg.getWillMessage().getBytes();
+//            ByteBuffer bb = (ByteBuffer) ByteBuffer.allocate(willPayload.length).put(willPayload).flip();
+//            PublishEvent pubEvt = new PublishEvent(msg.getWillTopic(), willQos,
+//                    bb, msg.isWillRetain(), msg.getClientID(), session);
+//            processPublish(pubEvt);
```
a5anka: Do we need to keep commented out code? We can get them back from GIT later if we want to.

PR #430, opened 2015-10-19
Fix DLC routing of topic messages

abeykoon:


sajinidesilva:
Is there a jira related to this?

PR #432, opened 2015-10-20
Consider durability in mqtt with regards to qos and clean session

akalankapagoda:
In MQTT subscribers, it is consider as durable where clean session is false and qos is greater than 0. Use this logic in locations where MQTT durability is in required.


modules/andes-core/broker/src/main/java/org/wso2/andes/mqtt/MQTTLocalSubscription.java
```diff
@@ -55,6 +55,9 @@
     //keep if the underlying subscription is active
     private boolean isActive;
 
+    // Should this subscriber act as a durable one
+    private boolean isDurable;
```
a5anka: Shouldn't we use following syntax for doc comments.
```java
/** 
 * Should this subscriber act as a durable one
 */
```

akalankapagoda: Yes, will convert these to doc comments.

PR #433, opened 2015-10-20
Fixed MB-1518 - Can consume queue messages through a topic exchange

sajinidesilva:


modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
```diff
@@ -472,15 +449,19 @@ private static void addLocalSubscriptionsForAllBindingsOfQueue(AMQQueue queue, S
...
+            List<Binding> validBindings = new ArrayList<Binding>();
             // Iterate unique bindings of the queue and add subscription entries.
             try {
                 for (Binding b : bindingList) {
                     // We ignore default exchange events. Andes doesn't honor creation of AMQ default exchange bindings
                     if (b.getExchange().getNameShortString().equals(ExchangeDefaults.DEFAULT_EXCHANGE_NAME.toString())) {
                         continue;
+                    } else{
+                        validBindings.add(b);
                     }
```
a5anka: I think we can just write this without an else. i.e.

```java
if (!b.getExchange().getNameShortString().equals(ExchangeDefaults.DEFAULT_EXCHANGE_NAME.toString())) {
   validBindings.add(b);
}
```

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/OrphanedSlotHandler.java
```diff
@@ -87,7 +87,8 @@ public void handleLocalSubscriptionsChanged(LocalSubscription subscription, Subs
                 if (null != matchingSubscription) {
                     reAssignSlotsIfNeeded(matchingSubscription);
                 } else {
-                    log.error("Deleting or disconnection a subscription which was not added previously");
+                    log.warn("Deleting or disconnection a subscription which was not added " +
+                            "previously");
```
a5anka: Do we need to break this in to two lines?

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
```diff
@@ -488,7 +469,11 @@ private static void addLocalSubscriptionsForAllBindingsOfQueue(AMQQueue queue, S
                         Andes.getInstance().openLocalSubscription(subscriptionEvent);
                         alreadyAddedSubscriptions.add(subscriptionEvent);
                     }
+                } else {
+                    throw new AndesException("Binding already exists for the queue " + queue
+                            .getName());
```
a5anka: Do we need to break this in to two lines?

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
```diff
@@ -59,11 +40,7 @@
 import org.wso2.andes.subscription.LocalSubscription;
 
 import java.nio.ByteBuffer;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
a5anka: We should not reduce to java.util.*

modules/andes-core/broker/src/main/java/org/wso2/andes/server/binding/BindingFactory.java
```diff
@@ -182,10 +182,15 @@ private synchronized boolean makeBinding(String bindingKey, AMQQueue queue, Exch
...
-            getConfigStore().addConfiguredObject(binding);
-            binding.logCreation();
+            try {
+                queue.addBinding(binding);
+                exchange.addBinding(binding);
+                getConfigStore().addConfiguredObject(binding);
+                binding.logCreation();
+            } catch (AndesException e) {
+                _bindings.remove(binding);
+                throw new AMQInternalException("Not permitted.", e);
```
a5anka: I think it better to have a more descriptive message

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
a5anka: I think it's  more appropriate to throw a AMQException. WDYT?

modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/SimpleAMQQueue.java
```diff
@@ -522,8 +523,34 @@ public void setDeleteOnNoConsumers(boolean b)
...
+            _logger.error("Binding already exists for the queue. Cannot be permitted");
+            throw new AndesException("Binding already exists for the queue ");
+        }
+    }
+
+    /**
+     * If binding is permitted, add it
+     * @param binding binding to be added
+     */
+    private void addBindingToQueue(final Binding binding){
```
a5anka: Do we need this method now?

modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/SimpleAMQQueue.java
```diff
@@ -522,8 +523,34 @@ public void setDeleteOnNoConsumers(boolean b)
...
+     * Add binding for the queue. Check if there are more than 1 binding other than the default
+     * binding. If so thrown an exception.
+     * @param binding  binding to be added
+     * @throws AndesException
+     */
+    public void addBinding(final Binding binding) throws AndesException {
+        if (_bindings.size() == 0) {
+            addBindingToQueue(binding);
+        } else if (_bindings.size() == 1) {
+            if (_bindings.get(0).getExchange().getNameShortString().equals(ExchangeDefaults
```
a5anka: Let's extract some local variables to make it more readable.

modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/SimpleAMQQueue.java
```diff
@@ -522,8 +523,34 @@ public void setDeleteOnNoConsumers(boolean b)
...
+                    .DEFAULT_EXCHANGE_NAME.toString())) {
+                addBindingToQueue(binding);
+            } else {
+                _logger.error("Binding already exists for the queue. Cannot be permitted");
+                throw new AndesException("Binding already exists for the queue");
```
a5anka: Do we need to do both logging and throwing?

modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/SimpleAMQQueue.java
```diff
@@ -522,8 +523,34 @@ public void setDeleteOnNoConsumers(boolean b)
...
+            }
+        } else if (_bindings.size() > 1) {
+            _logger.error("Binding already exists for the queue. Cannot be permitted");
+            throw new AndesException("Binding already exists for the queue ");
```
a5anka: Do we need to do both logging and throwing?

modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/SimpleAMQQueue.java
sajinidesilva: Yes. Because at the end of the day, this error is thrown only to client side. at MB side this error is not printed. Therefore error is printed here.

modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/SimpleAMQQueue.java
sajinidesilva: Please refer my previous reply

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
sajinidesilva: Agree

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
sajinidesilva: This is QpidBridge class. Also we throw this exception due to an limitation in andes code. This is not a violation of amqp spec. So I feel like throwing an Andes exception is better here. WDYT?

modules/andes-core/broker/src/main/java/org/wso2/andes/server/binding/BindingFactory.java
```diff
@@ -182,10 +182,15 @@ private synchronized boolean makeBinding(String bindingKey, AMQQueue queue, Exch
...
-            binding.logCreation();
+            try {
+                queue.addBinding(binding);
+                exchange.addBinding(binding);
+                getConfigStore().addConfiguredObject(binding);
+                binding.logCreation();
+            } catch (AndesException e) {
+                _bindings.remove(binding);
+                throw new AMQInternalException("Not permitted. Binding Already exists for this queue", e);
+            }
```
akalankapagoda: Can we throw a generic message here something like 'Adding binding failed'. 'e' will contain the real cause. Otherwise if any other exception occur here we assume everything as not permitted.

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
```diff
@@ -488,7 +490,10 @@ private static void addLocalSubscriptionsForAllBindingsOfQueue(AMQQueue queue, S
                         Andes.getInstance().openLocalSubscription(subscriptionEvent);
                         alreadyAddedSubscriptions.add(subscriptionEvent);
                     }
+                } else {
+                    throw new AndesException("Binding already exists for the queue " + queue.getName());
```
akalankapagoda: Just out of curiosity... :blush: what happens in a cluster? Are we considering a cluster when doing this validation?

modules/andes-core/broker/src/main/java/org/wso2/andes/server/binding/BindingFactory.java
sajinidesilva: the exception caught here is AndesException and Im giving tht exception only because of this reason actually.

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
a5anka: I think it will work in the cluster since we sync bind information in the cluster. But better to check.

modules/andes-core/broker/src/main/java/org/wso2/andes/server/binding/BindingFactory.java
a5anka: May be in the future there will be more. :confused:

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
a5anka: I don't think this is a limitation in Andes core, and it's better to have a clear separation between Andes core and transports. Everything before QpidBridge should be unaware of classes in Andes core and everything after QpidBridge should be unaware of transport specific classes. Only the bridge should be aware of classes in both sides. Therefore -1 in using AndesException.

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
sajinidesilva: This is valid for standalone as well as cluster

a5anka: Related to your previous reply.

> This is QpidBridge class. Also we throw this exception due to an limitation in andes code. This is not a violation  of amqp spec. So I feel like throwing an Andes exception is better here. WDYT?

I don't think this is a limitation in Andes core, and it's better to have a clear separation between Andes core and transports. Everything before QpidBridge should be unaware of classes in Andes core and everything after QpidBridge should be unaware of transport specific classes. Only the bridge should be aware of classes in both sides. Therefore -1 in using AndesException.

modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/SimpleAMQQueue.java
```diff
@@ -522,8 +523,34 @@ public void setDeleteOnNoConsumers(boolean b)
...
+     * binding. If so thrown an exception.
+     * @param binding  binding to be added
+     * @throws AndesException
+     */
+    public void addBinding(final Binding binding) throws AndesException {
+        if (_bindings.size() == 0) {
+            addBindingToQueue(binding);
+        } else if (_bindings.size() == 1) {
+            Binding existingBinding = _bindings.get(0);
+            if (existingBinding.getExchange().getNameShortString().equals(ExchangeDefaults.DEFAULT_EXCHANGE_NAME.toString())) {
```
a5anka: Let's extract a local variable to make it more readable.
e.g.
```java
bindingName = existingBinding.getExchange().getNameShortString()
```

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
sajinidesilva: (y)

modules/andes-core/broker/src/main/java/org/wso2/andes/server/binding/BindingFactory.java
ramith: use proper capitalization.

PR #435, opened 2015-10-21
Fixed MB-1518 - Can consume queue messages through a topic exchange

sajinidesilva:


modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
```diff
@@ -466,7 +466,8 @@ public static void removeBinding(Binding binding) throws AndesException {
      * @param subscription subscription
      * @throws AndesException
      */
-    private static void addLocalSubscriptionsForAllBindingsOfQueue(AMQQueue queue, Subscription subscription) throws AndesException, SubscriptionAlreadyExistsException {
+    private static void addLocalSubscriptionsForAllBindingsOfQueue(AMQQueue queue,
+                                                                   Subscription subscription) throws SubscriptionAlreadyExistsException, AMQException, AndesException {
```
a5anka: Do we need AndesException in this list now?

What about these locations? Can we change AndesException to AMQException.

https://github.com/wso2/andes/blob/master/modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/AMQQueue.java#L50

https://github.com/wso2/andes/blob/master/modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/SimpleAMQQueue.java#L541

They were added in https://github.com/wso2/andes/pull/433

sajinidesilva: We still throw AndesException because it throws another andes exception in this method. However this is a private method. This AndesException is caught  in the method where it is called which is also inside this class and throws AMQ exception. We do some stuff in the catch clause and that also throws andes exception. That is why it is handle that way.  HasithaH please correct me if I'm wrong

a5anka: I was able to remove AndesException by modifying catch block like below.

```java
            } catch (AndesException e) {
                log.warn("Reverting already created subscription entries for subscription " + subscription, e);
                for (InboundSubscriptionEvent alreadyAddedSub : alreadyAddedSubscriptions) {
                    try {
                        Andes.getInstance().closeLocalSubscription(alreadyAddedSub);
                    } catch (AndesException e1) {
                        // TODO: Check if we need to throw an AMQException or just logging is enough
                        log.warn("Error while closing local subscription", e1);
                    }
                }
                throw new AMQException("error while adding the local subscription", e);
            }
```

sajinidesilva: Yeah we could have done that. But this exception is catch in this same class and handling it there.

a5anka: If you want to pass it you can create an AMQException from the AndesException and pass?

PR #436, opened 2015-10-21
Modified debug log to info log in AMQProtocolHandler for client jar.

hemikak:
When the client is started with "amqj.protocol.logging.level=1" system property, all received packets to the client will be logged. But the sent packets cannot be seen as it is in "debug" level. 

Therefore changing it to "info" log.

### DIFF
--- a/modules/andes-core/client/src/main/java/org/wso2/andes/client/protocol/AMQProtocolHandler.java
+++ b/modules/andes-core/client/src/main/java/org/wso2/andes/client/protocol/AMQProtocolHandler.java
@@ -575,7 +575,7 @@ public class AMQProtocolHandler implements ProtocolEngine
         });
         if (PROTOCOL_DEBUG)
         {
-            _protocolLogger.debug(String.format("SEND: [%s] %s", this, frame));
+            _protocolLogger.info(String.format("SEND: [%s] %s", this, frame));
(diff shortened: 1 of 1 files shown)


ramith:
+1

a5anka:
:+1:

modules/andes-core/client/src/main/java/org/wso2/andes/client/protocol/AMQProtocolHandler.java
```diff
@@ -575,7 +575,7 @@ public void run()
         });
         if (PROTOCOL_DEBUG)
         {
-            _protocolLogger.debug(String.format("SEND: [%s] %s", this, frame));
+            _protocolLogger.info(String.format("SEND: [%s] %s", this, frame));
```
akalankapagoda: we're still checking PROTOCOL_DEBUG so wouldn't it print this only if PROTOCOL_DEBUG is set to true?

hemikak: logic is as follows :

bool PROTOCOL_DEBUG = sysprop(amqj.protocol.logging.level) != null.

its the same value as the systemprop.

PR #442, opened 2015-10-22
MB-1386, MB-1391, MB-1427, MB-1455, MB-1467, MB-1470

ThilankaBowala:
Changes related to invalid queue names/topic names/subscription ids

MB-1386 : Queue/Topic names should be trimmed before saving
MB-1391 : If a message is published from UI to a topic with a name which has a space in-between the message is not routed to the subscribers
MB-1427 : Prevent topic creation which has two full stops subsequently e.g - Sports..
MB-1455 : Queue names which has the " / " is not shown in UI
MB-1467 : Inconsistent behavior in super tenant and tenant space when trying to subscribe from JMS client
MB-1470 : "Cannot access the config registry" error message is displayed when trying to create a topic with *, # characters 

This fix includes validations of queue/topic names, when creating queues/topics from both UI and clients.

As said in the AMQP specification, this PR allows only alphanumeric, '.', '*' and '#' characters when creating topics. Also, prohibited space and the tenant separator '/', when creating queues.

Please merge with: https://github.com/wso2/carbon-business-messaging/pull/185

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/AMQQueueFactory.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/AMQQueueFactory.java
@@ -17,10 +17,12 @@
  */
 package org.wso2.andes.server.queue;
 
+import org.apache.log4j.Logger;
 import org.wso2.andes.AMQException;
 import org.wso2.andes.AMQSecurityException;
 import org.wso2.andes.framing.AMQShortString;
 import org.wso2.andes.framing.FieldTable;
+import org.wso2.andes.server.NameValidationUtils;
 import org.wso2.andes.server.virtualhost.VirtualHost;
 import org.wso2.andes.configuration.qpid.QueueConfiguration;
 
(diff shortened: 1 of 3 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/server/queue/AMQQueueFactory.java
```diff
@@ -148,6 +151,28 @@ public static AMQQueue createAMQQueueImpl(String queueName,
...
+        if(_logger.isDebugEnabled()){
+            _logger.debug("Queue name is: " + queueName + ", durable=" + durable + ", owner: " + owner);
+        }
+
+        // Prevent creating queue, when queue name/subscription id is invalid
+        if(!NameValidationUtils.isValidQueueName(queueName)){
+            virtualHost.getQueueRegistry().unregisterQueue(new AMQShortString(queueName));
+
+            String queueTypeIndicator;
+            if(durable == true) {
```
sdkottegoda: Let's flip this to true == durable

PR #443, opened 2015-10-22
Revert "Fixed MB-1518 - Can consume queue messages through a topic ex…

a5anka:
Reverts wso2/andes#435


ramith:
I reverted https://wso2.org/jira/browse/MB-1518

a5anka:
This tracked in PR 444

PR #449, opened 2015-10-23
[MB-1482] Fixing JMS Transaction flow

hastef88:
**Expectation from a session-transacted consumer** : 

1. Say we publish messages 1,2,3,4,5,6,7,8,9,10 to a queue.  
2. Consumer performs a session.commit upon receiving message 5. (messages 1,2,3,4,5 should be acked and cleared from the server.)
3. Consumer performs a session.rollback at message 7. (messages 6,7 should be rejected back to the server for a requeue.) 
4. Consumer should receive message 6,7 until it decides to send a session.commit. 
5. Consumer performs a session.commit at message 7. (messages 6,7 are cleared and consumer should receive message 8,9,10 ). 

Refer [1] for more details. 

**Issues fixed in JMS Transaction flow** : 

(1) 

Andes client maintains an internal BlockingQueue to manage the inbound messages before exposing them to the consuming application. At the time of rollback (e.g. message 7), this buffer can already have messages beyond the rollback point (e.g 8,9,10). At a rollback event, these messages are also rejected in order to keep consistency. 

Since these messages are not yet exposed to the end-user, their hidden re-delivery attempts will accumulate, causing them to go DLC sooner than expected. (exceeding maxDeliveryCount). To avoid this, we no longer increment deliveryCount for messages beyond the last rollbacked message ID.

(2) 

The messages were being rejected in the wrong order. (Messages 8,9,10 being rejected before 6 and 7.)

(3) 

Use of "synchronized" does not guarantee that the longest-waiting thread gets the next execution. (This could mess up the message order in a continous rollback scenario). Therefore, switched 2 locks within andes client to "ReentrantLock"s with fair-policy = true. Refer [2] and [3] for more details. 

**Pending Tasks** : 

Cleaning the PR
Integration test cases for transactions
Further tests in cluster

[1] : http://www.javaworld.com/article/2074123/java-web-development/transaction-and-redelivery-in-jms.html?page=2
[2] : http://stackoverflow.com/questions/11821801/why-use-a-reentrantlock-if-one-can-use-synchronizedthis
[3] : http://stackoverflow.com/questions/12578687/waiting-on-a-condition-in-a-reentrant-lock

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
@@ -258,6 +258,10 @@ public class QpidAndesBridge {
                     getSubscriptionStore().getLocalSubscriptionForChannelId(channel.getId());
             DeliverableAndesMetadata rejectedMessage = localSubscription.getMessageByMessageID(message.getMessageId());
 
+            channel.setLastRejectedMessageId(message.getMessageNumber());
+
+            rejectedMessage.setIsBeyondLastRollbackedMessage(channel.isMessageBeyondLastRollback(message.getMessageId()));
+
             log.debug("AMQP BRIDGE: rejected message id= " + rejectedMessage.getMessageID()
                     + " channel = " + channel.getId());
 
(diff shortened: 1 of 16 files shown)


ramith:
Merge this on monday ( post beta)

PR #457, opened 2015-10-30
Format code by removing unneccessary clear commands

sajinidesilva:


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/disruptor/inbound/MessagePreProcessor.java
```diff
@@ -226,18 +226,8 @@ private void handleTopicRoutine(InboundEventContainer event, AndesMessage messag
...
-                    event.clearMessageList();
-                } else {
-                    // Since no matching subscribers and not a retained enabled message, Event can be
-                    // cleared and set to ignore the message by message writers
-                    event.clear();
-                    log.info("Message routing key: " + message.getMetadata().getDestination() + " No routes in " +
-                             "cluster. Ignoring Message id " + message.getMetadata().getMessageID());
-                }
+                log.info("Message routing key: " + message.getMetadata().getDestination() + " No routes in " +
+                        "cluster. Ignoring Message id " + message.getMetadata().getMessageID());
```
Asitha: Don't we have retain now? Is it moved to somewhere else?

PR #459, opened 2015-10-30
Initialize current node's last published ID before recovering slots

a5anka:


hemikak:
JIRA - https://wso2.org/jira/browse/MB-1547

PR #462, opened 2015-11-02
[MB-1533] Added check to find active subscribers when deleting durabe subscriptions

sdkottegoda:
Addresses the issue reported at https://wso2.org/jira/browse/MB-1533

The issue occurred since the the check for active subscriptions upon a queue delete request was only done for queues but not for topics.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/SubscriptionStore.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/SubscriptionStore.java
@@ -330,24 +330,64 @@ public class SubscriptionStore {
 
     /**
      * get all (active/inactive) CLUSTER subscriptions for a queue/topic.
+     * <p/>
+     * If it is for a Queue, the destination will be the name of the queue. If it is for a topic, the destination is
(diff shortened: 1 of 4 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/SubscriptionStore.java
```diff
@@ -333,24 +333,61 @@ public int numberOfSubscriptionsInCluster(String destination, boolean isTopic, S
...
+                    SubscriptionType.AMQP);
+            for (AndesSubscription sub : topicSubsOfDest) {
+                if (sub.getTargetQueue().equals(queueName)) {
+                    subscriptionsOfQueue.add(sub);
+                }
+            }
+            //Get all MQTT subscriptions
+            topicSubsOfDest = getAllSubscribersForDestination(destination, true, SubscriptionType.MQTT);
+            for (AndesSubscription sub : topicSubsOfDest) {
+                if (sub.getTargetQueue().equals(queueName)) {
```
sajinidesilva: Shall we flip this statement?

PR #465, opened 2015-11-02
[MB-1545] Prohibited deletion of DeadLetterChannel

sdkottegoda:
Addresses the issue reported at https://wso2.org/jira/browse/MB-1545

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/server/handler/QueueDeleteHandler.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/server/handler/QueueDeleteHandler.java
@@ -30,6 +30,7 @@ import org.wso2.andes.server.ClusterResourceHolder;
 import org.wso2.andes.server.protocol.AMQProtocolSession;
 import org.wso2.andes.server.protocol.AMQSessionModel;
 import org.wso2.andes.server.queue.AMQQueue;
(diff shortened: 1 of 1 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/server/handler/QueueDeleteHandler.java
```diff
@@ -85,60 +86,53 @@ public void methodReceived(AMQStateManager stateManager, QueueDeleteBody body, i
...
-                    protocolConnection.writeFrame(responseBody.generateFrame(channelId));
-                }
-                else
-                {
-                    _logger.warn("Cannot Delete Queue" + queue.getName() + " It Has Registered Subscriptions.");
-                    throw new AMQException("Cannot Delete Queue" + queue.getName() + " It Has Registered Subscriptions.");
                 }
             }
+        } else {
+            _logger.warn("Deletion of Dead letter channel not permitted.");
```
sajinidesilva: Shouldn't this be an error log?

sdkottegoda: From the server's side, is it an error? I thought it should be a warning saying somebody attempted to delete the DLC and it is not permitted.

PR #468, opened 2015-11-03
MQTT close previous connection if a new client joined with the same client id

akalankapagoda:
This pull fixes the issue https://wso2.org/jira/browse/MB-1548.

When a new client joins with a client id that is the same as an already connected client, according to MQTT spec, the existing client should be disconnected and the new client should be connected.

http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718032

"If the ClientId represents a Client already connected to the Server then the Server MUST disconnect the existing Client [MQTT-3.1.4-2]."

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/dna/mqtt/moquette/server/ServerChannel.java
+++ b/modules/andes-core/broker/src/main/java/org/dna/mqtt/moquette/server/ServerChannel.java
@@ -1,6 +1,9 @@
 package org.dna.mqtt.moquette.server;
 
 import io.netty.channel.Channel;
+
+import java.util.UUID;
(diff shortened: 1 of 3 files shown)


modules/andes-core/broker/src/main/java/org/dna/mqtt/moquette/server/ServerChannel.java
```diff
@@ -1,6 +1,9 @@
 package org.dna.mqtt.moquette.server;
```
sajinidesilva: Should add the copyright header here

PR #473, opened 2015-11-05
Fixing MB-1559

hastef88:
Initial commit for calculating local safe zone based on minimum ID across all destinations of a single node.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotCoordinator.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotCoordinator.java
@@ -36,6 +36,7 @@ public interface SlotCoordinator {
      * @param queueName  Name of the queue
      * @param startMessageId Start of message ID of the slot
      * @param endMessageId End message ID of the slot
+     * @param localSafeZone
      * @throws ConnectionException
      */
     public void updateMessageId(String queueName,long startMessageId, long endMessageId, long localSafeZone) throws ConnectionException;
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotCoordinator.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotCoordinator.java
@@ -38,7 +38,7 @@ public interface SlotCoordinator {
      * @param endMessageId End message ID of the slot
      * @throws ConnectionException
      */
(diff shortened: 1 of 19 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/slot/gen/SlotManagementService.java
```diff
@@ -26,7 +26,7 @@
```
hastef88: Auto generated file from thrift-gen based on slot.thrift file.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotCoordinator.java
```diff
@@ -38,7 +38,7 @@
      * @param endMessageId End message ID of the slot
      * @throws ConnectionException
      */
-    public void updateMessageId(String queueName,long startMessageId, long endMessageId) throws ConnectionException;
+    public void updateMessageId(String queueName,long startMessageId, long endMessageId, long localSafeZone) throws ConnectionException;
```
sajinidesilva: Update the comment for the new parameter

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotManagerClusterMode.java
```diff
@@ -430,7 +432,7 @@ public void reAssignSlotWhenNoSubscribers(String nodeId, String queueName) throw
 		}
 	}
 
-	protected Long getLastPublishedIDByNode(String nodeID) throws AndesException {
+	protected Long getLocalSafeZone(String nodeID) throws AndesException {
```
sajinidesilva: Is it correct to return the last published ID as the local safe zone?

hastef88: nope. this is a naming confusion since the old method is re-used. will refactor.

hastef88: fixed

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotCoordinator.java
hastef88: fixed

PR #476, opened 2015-11-06
Slots are retained in the node until global safe zone exeeds slot range

a5anka:


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesContextStore.java
```diff
@@ -261,7 +261,7 @@ void createSlot(long startMessageId, long endMessageId, String storageQueueName,
      * @param endMessageId end message id of slot
      * @throws AndesException
      */
```
abeykoon: Doc comment is not updated

PR #479, opened 2015-12-03
Delete Queue mapping

sdkottegoda:
Applied the changes in https://github.com/wso2/andes/pull/377 with fixes for deleting the queue mapping entries from in-memory maps upon the receipt of a cluster queue delete notification.

Also, the Hashmap for storing queue mapping was removed and a guava cache was used in place.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AMQPConstructStore.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AMQPConstructStore.java
@@ -122,16 +122,27 @@ public class AMQPConstructStore {
      * remove a queue
      *
      * @param queueName name of the queue to be removed
-     * @param isLocal   is this a local change
      * @throws AndesException
      */
-    public void removeQueue(String queueName, boolean isLocal) throws AndesException {
-        if (isLocal) {
-            andesContextStore.deleteQueueInformation(queueName);
-            //create the space created to keep message counter on this queue
-            messageStore.removeQueue(queueName);
-        }
+    public void removeQueue(String queueName) throws AndesException {
+
+        // Remove the queue from internal maps
+        removeLocalQueueData(queueName);
+
+        // Remove queue information from database
+        andesContextStore.deleteQueueInformation(queueName);
+        messageStore.removeQueue(queueName);
+    }
+
+    /**
+     * remove a queue from local maps
+     *
+     * @param queueName name of the queue to be removed
+     * @throws AndesException
+     */
+    public void removeLocalQueueData(String queueName) throws AndesException {
         andesQueues.remove(queueName);
+        messageStore.removeLocalQueueData(queueName);
     }
 
     /**
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessageStore.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessageStore.java
@@ -68,44 +68,10 @@ public interface MessageStore extends HealthAwareStore{
     Map<Long, List<AndesMessagePart>> getContent(List<Long> messageIDList) throws AndesException;
 
     /**
-     * store mata data of messages
-     *
-     * @param metadataList metadata list to store
-     * @throws AndesException
-     */
-    void addMetadata(List<AndesMessageMetadata> metadataList) throws AndesException;
-
-    /**
-     * store metadata of a single message
-     *
-     * @param metadata metadata to store
-     * @throws AndesException
-     */
-    void addMetadata(AndesMessageMetadata metadata) throws AndesException;
-
-    /**
      * Store messages into database.
      * @param messageList messages to be stored
      */
     void storeMessages(List<AndesMessage> messageList) throws AndesException;
-
-    /**
-     * store metadata specifically under a queue
-     *
-     * @param queueName name of the queue to store metadata
-     * @param metadata metadata to store
-     * @throws AndesException
-     */
-    void addMetadataToQueue(final String queueName, AndesMessageMetadata metadata) throws AndesException;
-
-    /**
-     * store metadata list specifically under a queue
-     *
-     * @param queueName name of the queue to store metadata
-     * @param metadata metadata list to store
-     * @throws AndesException
-     */
-    void addMetadataToQueue(final String queueName, List<AndesMessageMetadata> metadata) throws AndesException;
 
     /**
      * Store a message in a different Queue without altering the meta data.
@@ -358,6 +324,13 @@ public interface MessageStore extends HealthAwareStore{
      * @param storageQueueName name of the queue actually stored in DB
      */
     void removeQueue(String storageQueueName) throws AndesException;
+
+    /**
+     * Remove queue entry from the queue mapping cache
+     *
+     * @param storageQueueName the name of the queue to be removed
+     */
+    void removeLocalQueueData(String storageQueueName);
 
     /**
      * Increment message counter for a queue by a given incrementBy value
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/store/FailureObservingMessageStore.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/store/FailureObservingMessageStore.java
@@ -116,62 +116,10 @@ public class FailureObservingMessageStore implements MessageStore {
         }
     }
 
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public void addMetadata(List<AndesMessageMetadata> metadataList) throws AndesException {
-        try {
-            wrappedInstance.addMetadata(metadataList);
-        } catch (AndesStoreUnavailableException exception) {
-            notifyFailures(exception);
-            throw exception;
-        }
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public void addMetadata(AndesMessageMetadata metadata) throws AndesException {
-        try {
-            wrappedInstance.addMetadata(metadata);
-        } catch (AndesStoreUnavailableException exception) {
-            notifyFailures(exception);
-            throw exception;
-        }
-    }
-
     @Override
     public void storeMessages(List<AndesMessage> messageList) throws AndesException {
         try {
             wrappedInstance.storeMessages(messageList);
-        } catch (AndesStoreUnavailableException exception) {
-            notifyFailures(exception);
-            throw exception;
-        }
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public void addMetadataToQueue(String queueName, AndesMessageMetadata metadata) throws AndesException {
-        try {
-            wrappedInstance.addMetadataToQueue(queueName, metadata);
-        } catch (AndesStoreUnavailableException exception) {
-            notifyFailures(exception);
-            throw exception;
-        }
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public void addMetadataToQueue(String queueName, List<AndesMessageMetadata> metadata) throws AndesException {
-        try {
-            wrappedInstance.addMetadataToQueue(queueName, metadata);
         } catch (AndesStoreUnavailableException exception) {
             notifyFailures(exception);
             throw exception;
@@ -525,6 +473,14 @@ public class FailureObservingMessageStore implements MessageStore {
             notifyFailures(exception);
             throw exception;
         }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void removeLocalQueueData(String storageQueueName) {
+        wrappedInstance.removeLocalQueueData(storageQueueName);
     }
 
     /**
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
@@ -18,6 +18,10 @@
 
 package org.wso2.andes.store.rdbms;
 
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import com.google.common.cache.Weigher;
 import org.apache.log4j.Logger;
 import org.wso2.andes.configuration.util.ConfigurationProperties;
 import org.wso2.andes.kernel.AndesContextStore;
@@ -48,7 +52,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 
 import static org.wso2.andes.store.rdbms.RDBMSConstants.CONTENT_TABLE;
 import static org.wso2.andes.store.rdbms.RDBMSConstants.MESSAGE_CONTENT;
@@ -70,7 +74,6 @@ public class RDBMSMessageStoreImpl implements MessageStore {
     /**
      * Cache queue name to queue_id mapping to avoid extra sql queries
      */
-    private final Map<String, Integer> queueMap;
 
     private RDBMSConnection rdbmsConnection;
 
@@ -93,10 +96,11 @@ public class RDBMSMessageStoreImpl implements MessageStore {
                     " FROM " + CONTENT_TABLE +
                     " WHERE " + MESSAGE_ID + " IN (";
 
-    public RDBMSMessageStoreImpl() {
-        queueMap = new ConcurrentHashMap<>();
-        
-    }
+    /**
+     * The cache which holds the queue mappings(queue name to queue id) in memory
+     * In the absence of a queried queue name in the cache, the queue id is loaded from the database
+     */
+    private LoadingCache<String, Integer> queueMappings;
 
     /**
      * {@inheritDoc}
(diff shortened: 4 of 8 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
```diff
@@ -112,11 +114,34 @@ public DurableStoreConnection initializeMessageStore(AndesContextStore contextSt
...
+    private void initializeQueueMappingCache() {
+        queueIDs = CacheBuilder.newBuilder()
+                .concurrencyLevel(5)
+                .maximumSize(1000)
+                .build(
+                        new CacheLoader<String, Integer>() {
+                            public Integer load(String queueName) throws AndesException {
+                                try {
+                                    Integer queueID = getQueueID(queueName);
+                                    log.debug("Loaded queue: " + queueName + " to the cache from database");
```
sajinidesilva: Shall we check whether debugging is enabled?

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
```diff
@@ -112,11 +114,34 @@ public DurableStoreConnection initializeMessageStore(AndesContextStore contextSt
...
+                                    return queueID;
+                                } catch (SQLException e) {
+                                    throw new AndesException("Error retrieving queue id for queue: " + queueName);
```
sajinidesilva: Shall we pass the exception as well?

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
```diff
@@ -1279,28 +1322,18 @@ public void deleteMessagesFromExpiryQueue(List<Long> messagesToRemove) throws An
...
-        // If queue is not available create a queue in DB
-        int queueID = getQueueID(destinationQueueName);
-
-        if (queueID != -1) {
-            queueMap.put(destinationQueueName, queueID);
+    private int getCachedQueueID(final String destinationQueueName) throws AndesException {
+        try {
+            return queueIDs.get(destinationQueueName);
+        } catch (ExecutionException e) {
+            throw new AndesException("Error retrieving queue id for queue: " + destinationQueueName + " from cache");
```
sajinidesilva: Shall we pass the exception e along with AndesException?

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
```diff
@@ -1782,7 +1815,39 @@ public void resetMessageCounterForQueue(String storageQueueName) throws AndesExc
...
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void removeCachedQueue(String storageQueueName) {
+        queueIDs.invalidate(storageQueueName);
+        log.debug("Queue: " + storageQueueName + " removed from cache.");
```
sajinidesilva: Is there any reason why we haven't check whether debugging is enabled?

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
```diff
@@ -700,23 +742,24 @@ public void updateMetadataInformation(String currentQueueName, List<AndesMessage
...
      * @throws SQLException
      */
-    private void addMetadataToBatch(PreparedStatement preparedStatement,
-                                    AndesMessageMetadata metadata,
-                                    final String queueName) throws SQLException {
+    private void addMetadataToBatch(PreparedStatement preparedStatement, AndesMessageMetadata metadata,
+                                    final String queueName) throws AndesException {
 
-        Context metaAdditionToBatchContext = MetricManager.timer(Level.INFO, MetricsConstants.ADD_META_DATA_TO_BATCH).start();
+        Context metaAdditionToBatchContext = MetricManager.timer(Level.INFO, MetricsConstants.ADD_META_DATA_TO_BATCH)
```
a5anka: @AnujaLK 
metaAdditionToBatchContext and contextWrite are used for the same purpose. Why do we need two? Can't we use just one?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessageStore.java
```diff
@@ -360,6 +360,11 @@ void addMessageToExpiryQueue(Long messageId, Long expirationTime, boolean isMess
     void removeQueue(String storageQueueName) throws AndesException;
 
     /**
+     * Remove queue entry from the queue mapping cache
+     */
```
sajinidesilva: Its good to define parameters also in the comment

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
```diff
@@ -112,11 +114,34 @@ public DurableStoreConnection initializeMessageStore(AndesContextStore contextSt
...
         return rdbmsConnection;
     }
 
     /**
+     * Method to initialize the queue mapping cache.
+     */
+    private void initializeQueueMappingCache() {
+        queueIDs = CacheBuilder.newBuilder()
+                .concurrencyLevel(5)
+                .maximumSize(1000)
```
a5anka: Better to make these constants? maximumSize should be a tuning paramater IMO.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
a5anka: Please ignore this. I understand now.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/FailureObservingMessageStore.java
```diff
@@ -531,6 +531,14 @@ public void removeQueue(String storageQueueName) throws AndesException {
      * {@inheritDoc}
      */
     @Override
+    public void removeCachedQueue(String storageQueueName) {
```
ramith: Conceptually caching is internal to message store isn't? 
Is there a possibility handle this with in the message store impl?

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
ramith: +1

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AMQPConstructStore.java
```diff
@@ -126,7 +126,13 @@ public void addQueue(AndesQueue queue, boolean isLocal) throws AndesException {
...
      */
     public void removeQueue(String queueName, boolean isLocal) throws AndesException {
-        if (isLocal) {
+
+        // If the queue deletion call was originally received by this node, the queue should be removed from the
+        // database as well. Else, only the entry of the queue mapping cache should be removed.
+        if (!isLocal){
+            messageStore.removeCachedQueue(queueName);
+        }
+        else {
```
a5anka: Incorrect line break

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AMQPConstructStore.java
```diff
@@ -126,7 +126,13 @@ public void addQueue(AndesQueue queue, boolean isLocal) throws AndesException {
      * @throws AndesException
      */
     public void removeQueue(String queueName, boolean isLocal) throws AndesException {
-        if (isLocal) {
+
+        // If the queue deletion call was originally received by this node, the queue should be removed from the
+        // database as well. Else, only the entry of the queue mapping cache should be removed.
+        if (!isLocal){
+            messageStore.removeCachedQueue(queueName);
```
a5anka: I don't this we have to pass "isLocal" and at runtime decide on the pathe we should take. We know if isLocal true or false at the compile time. Therefore we should have used two methods for this. WDYT?

sajinidesilva: +1

a5anka: I think we should also invalidate the local nodes cache too.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
```diff
@@ -405,8 +433,13 @@ public void addMetadata(AndesMessageMetadata metadata) throws AndesException {
            rollback(connection, RDBMSConstants.TASK_ADDING_METADATA);
            throw new AndesException("Error occurred while inserting metadata list to queues ", bue);
             
-        }*/ catch (SQLException e) {
+        }*/ catch (AndesException e) {
```
a5anka: Let's get rid of the commented out code.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
```diff
@@ -405,8 +433,13 @@ public void addMetadata(AndesMessageMetadata metadata) throws AndesException {
...
+            rollback(connection, RDBMSConstants.TASK_ADDING_METADATA);
+            throw e;
+        }catch (SQLException e){
```
malakaganga: We need to have a space between "}" and "catch" I guess

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
```diff
@@ -494,7 +529,11 @@ public void addMetadataToQueue(String queueName, AndesMessageMetadata metadata)
            rollback(connection, RDBMSConstants.TASK_ADDING_METADATA_TO_QUEUE + queueName);
            throw new AndesException("Error occurred while inserting message metadata to queue ", bue);
             
-        }*/ catch (SQLException e) {
+        }*/
+        catch (AndesException e) {
```
a5anka: Let's get rid of the commented out code and fix line break issue.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
sdkottegoda: Made them constants. But I do not think we need a configuration for the maximum size. 

A size of 1MB is enough to store more than 10,000 queue names. By making the size 2MB, we can assure that the message cache will have more than enough space to hold all the queue mappings and 2MB is not that big of a size for the memory.

Do you think we still need a configuration for this?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AMQPConstructStore.java
```diff
@@ -122,15 +122,26 @@ public void addQueue(AndesQueue queue, boolean isLocal) throws AndesException {
...
-    public void removeQueue(String queueName, boolean isLocal) throws AndesException {
-        if (isLocal) {
-            andesContextStore.deleteQueueInformation(queueName);
-            //create the space created to keep message counter on this queue
-            messageStore.removeQueue(queueName);
-        }
+    public void removeLocalQueue(String queueName) throws AndesException {
+
+        // Remove the queue from internal maps
+        removeClusterQueue(queueName);
```
a5anka: I think it's better to invalidate the cache after removing the entry from DB, to avoid any race conditions. Since the reads can happen concurrently this can happen in a loaded environment.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AMQPConstructStore.java
```diff
@@ -122,15 +122,26 @@ public void addQueue(AndesQueue queue, boolean isLocal) throws AndesException {
...
+        messageStore.removeQueue(queueName);
+    }
+
+    /**
+     * remove a queue from local maps
+     *
+     * @param queueName name of the queue to be removed
+     * @throws AndesException
+     */
+    public void removeClusterQueue(String queueName) throws AndesException {
```
a5anka: Shall we change method name to something like clearQueueState? removeClusterQueue does not seem to describe the task it's doing (remove a queue from local maps) correctly

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
```diff
@@ -864,8 +891,8 @@ public AndesMessageMetadata getMetadata(long messageId) throws AndesException {
             }
             if (log.isDebugEnabled()) {
                 log.debug("request: metadata range (" + firstMsgId + " , " + lastMsgID +
-                        ") in destination queue " + storageQueueName
-                        + "\nresponse: metadata count " + metadataList.size());
+                          ") in destination queue " + storageQueueName
```
a5anka: sentence should start with '+'

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
a5anka: Ok. No problem. In the worst case it will read entries from DB and the memory foot print is very low.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AMQPConstructStore.java
sdkottegoda: I'm sorry I don't get how invalidating the cache afterwards could avoid race conditions.

The rationale behind removing it beforehand was that the queue removal should be reflected in the memory before it gets removed from the store, or else if somebody tries to publish a message to a queue which is deleted from the db but not yet from the memory, there will be problems.

a5anka: Similarly after invalidating the cache before removing the entry some can read the cache. When that happens the old queue mapping is reloaded and will be used even after the db entry is deleted. Therefore it seems changing the order won't solve the issue.

Alternative approaches I can think of are,

1. Use a read write lock to make the operations atomic
2. Use a hash instead of an auto incremented id for queue mapping.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
```diff
@@ -93,10 +96,10 @@
...
                     " WHERE " + MESSAGE_ID + " IN (";
 
-    public RDBMSMessageStoreImpl() {
-        queueMap = new ConcurrentHashMap<>();
-        
-    }
+    /**
+     * The cache which holds the queue mappings(queue name to queue id) in memory
+     */
+    private LoadingCache<String, Integer> queueIDs;
```
abeykoon: Shall we update the variable name? It can indicate loading what.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
```diff
@@ -444,6 +475,20 @@ public void storeMessages(List<AndesMessage> messageList) throws AndesException
             storeMetadataPS.executeBatch();
             storeContentPS.executeBatch();
             connection.commit();
+
+            // Add messages to cache after adding them to the database
+            // Messages are added afterwards since we need to add messages to the cache only if they are added to the
+            // database.
+            addToCache(messageList);
+        } catch (BatchUpdateException bue) {
+
```
abeykoon: unnecessary line space

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
```diff
@@ -863,13 +948,13 @@ public AndesMessageMetadata getMetadata(long messageId) throws AndesException {
                 MessageTracer.trace(md, MessageTracer.METADATA_READ_FROM_DB + " slot = " + slot.getId());
             }
             if (log.isDebugEnabled()) {
-                log.debug("request: metadata range (" + firstMsgId + " , " + lastMsgID +
-                        ") in destination queue " + storageQueueName
-                        + "\nresponse: metadata count " + metadataList.size());
+                log.debug("request: metadata range (" + firstMsgId + " , " + lastMsgID + ") in destination queue "
+                          + storageQueueName + "\nresponse: metadata count " + metadataList.size());
```
abeykoon: why \n here? A new line can complicate things when concurrency is there.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
```diff
@@ -112,11 +115,49 @@ public DurableStoreConnection initializeMessageStore(AndesContextStore contextSt
...
         
         this.messageCache = (new MessageCacheFactory()).create();
+        initializeQueueMappingCache();
+
         log.info("Message Store initialised");
         return rdbmsConnection;
     }
 
     /**
+     * Method to initialize the queue mapping cache.
```
abeykoon: Let us say what queue mapping cache does in a sentence. It adds more clarity.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
a5anka: :+1:

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
sdkottegoda: The variable name 'queueIds' says what it holds. Updated the name to 'queueMappings' and updated the comment to indicate how data is loaded from the database in the absence of a queue name.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
sdkottegoda: added a ', ' instead

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
sdkottegoda: done

PR #480, opened 2015-12-03
Add MBean for subscription deletion

a5anka:


modules/andes-core/broker/src/main/java/org/wso2/andes/mqtt/MQTTLocalSubscription.java
```diff
@@ -147,6 +149,14 @@ public void setMqqtServerChannel(MQTTopicManager mqqtServerChannel) {
      * {@inheritDoc}
      */
     @Override
+    public void forcefullyDisconnect() throws AndesException{
```
malakaganga: We should have a space between "AndesException" and "{" I guess...

a5anka: :+1:

PR #485, opened 2015-12-14
Remove message iterator from delivery strategies

a5anka:


ramith:
@a5anka  are we going to merge this one?

akalankapagoda:
Since sorted sets add an overhead to the load while sorting we're looking for a data structure which can handle concurrency with the iterators.

a5anka:
:+1:

PR #418, opened 2015-10-14
[MB-1433] Create queue when publishing

sdkottegoda:
Addresses the issue reported at https://wso2.org/jira/browse/MB-1433

Added queues and bindings when creating a publisher.

PR #486, opened 2015-12-16
Publish to a non existing queue

sdkottegoda:
Applied the changes in #377 with fixes for deleting the queue mapping entries from in-memory maps upon the receipt of a cluster queue delete notification.

Also, the Hashmap for storing queue mapping was removed and a guava cache was used in place.

In addition, https://github.com/wso2/andes/pull/418 ([MB-1433] Create queue when publishing) was reverted and a message published to a non-existing queue was dropped.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/mqtt/connectors/PersistenceStoreConnector.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/mqtt/connectors/PersistenceStoreConnector.java
@@ -101,15 +101,15 @@ public class PersistenceStoreConnector implements MQTTConnector {
                 publisher = new MQTTPublisherChannel(messageContext.getChannel());
                 publisherTopicCorrelate.put(messageContext.getPublisherID(), publisher);
                 //Finally will register the publisher channel for flow controlling
-                AndesChannel publisherChannel = Andes.getInstance().createChannel(publisher);
+
+                String andesChannelId = MQTTUtils.DEFAULT_ANDES_CHANNEL_IDENTIFIER;
+                if (null != messageContext.getChannel()) {
+                    andesChannelId = messageContext.getChannel().remoteAddress().toString().substring(1);
+                }
+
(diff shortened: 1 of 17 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/mqtt/connectors/PersistenceStoreConnector.java
```diff
@@ -101,15 +101,15 @@ public void addMessage(MQTTMessageContext messageContext) throws MQTTException {
                 publisher = new MQTTPublisherChannel(messageContext.getChannel());
                 publisherTopicCorrelate.put(messageContext.getPublisherID(), publisher);
                 //Finally will register the publisher channel for flow controlling
-                AndesChannel publisherChannel = Andes.getInstance().createChannel(publisher);
-                //Set channel details
-                //Substring to remove leading slash character from remote address
+
+                String andesChannelId = "MQTT-Unknown";
```
sajinidesilva: Shall we make this String to a static String variable?

modules/andes-core/broker/src/main/java/org/wso2/andes/server/AMQChannel.java
```diff
@@ -385,77 +395,71 @@ private void deliverCurrentMessageIfComplete()
...
-                            if(!AndesContext.getInstance().isClusteringEnabled()) {
-                                _logger.warn("MESSAGE DISCARDED: No routes for message - " + createAMQMessage(_currentMessage));
-                            }
+                    // If the destination of the andes channel is not set, set the destination to the routing key to the
+                    // receive message
+                    if (null == andesChannel.getDestination()) {
+                        andesChannel.setDestination(this._currentMessage.getRoutingKey());
+                    }
+
+                    if ((null == destinationQueues || _currentMessage.getDestinationQueues().isEmpty())
```
sajinidesilva: Its more readable if you separate two statements inside if condition with parenthesis.

PR #497, opened 2016-01-04
Changing Subscription Store to support different protocol types

akalankapagoda:
This will separate subscriptions kept in memory for each protocol type and give the responsibility of adding, updating and removing those subscriptions to a protocol specific implementation of an interface.


modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPClusterSubscriptionHandler.java
```diff
@@ -0,0 +1,211 @@
+package org.wso2.andes.amqp;
+
+import org.wso2.andes.kernel.AndesException;
+import org.wso2.andes.kernel.AndesSubscription;
+import org.wso2.andes.kernel.DestinationType;
+import org.wso2.andes.kernel.ProtocolType;
+import org.wso2.andes.subscription.SubscriptionHandler;
+import org.wso2.andes.subscription.TopicSubscriptionBitMapHandler;
+
+import java.util.*;
```
hemikak: Have to expand.

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPClusterSubscriptionHandler.java
```diff
@@ -0,0 +1,211 @@
...
+
+public class AMQPClusterSubscriptionHandler implements SubscriptionHandler {
+
+
+    /**
+     * All topic subscriptions will be handled through this.
+     */
+    SubscriptionHandler topicSubscriptionHandler;
+
+    SubscriptionHandler durableTopicSubscriptionHandler;
```
hastef88: Comment needed

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPClusterSubscriptionHandler.java
```diff
@@ -0,0 +1,211 @@
...
+
+    /**
+     * All queue subscriptions will go into this.
+     */
+    private Map<String, Set<AndesSubscription>> queueSubscriptionMap = new ConcurrentHashMap<>();
+
+    public AMQPClusterSubscriptionHandler() throws AndesException {
+        topicSubscriptionHandler = new TopicSubscriptionBitMapHandler(ProtocolType.AMQP);
+        durableTopicSubscriptionHandler = new TopicSubscriptionBitMapHandler(ProtocolType.AMQP);
```
hastef88: Do we use this handler for queues as well (durable topics and queues both) ?

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPClusterSubscriptionHandler.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Set<String> getAllDestinations(DestinationType destinationType) {
+        Set<String> destinations;
+
+        if (DestinationType.TOPIC == destinationType) {
```
sajinidesilva: Rather than checking whether it is topic or durable topic always in each method, cant we decide the subscription path in an upper level? So that topic subscription goes in one path and durable subscriptions goes in another path?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/StorageQueueData.java
```diff
@@ -0,0 +1,38 @@
+package org.wso2.andes.kernel.slot;
```
hastef88: License + class and variable comments needed

modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/BasicSubscription.java
```diff
@@ -77,8 +80,6 @@ public BasicSubscription(String subscriptionAsStr) {
                 this.subscriptionID = tokens[1];
```
hastef88: It would be more readable if the split string is put into variable before using as below (like String key = tokens[0]; )

modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/LocalDurableTopicSubscriptionStore.java
```diff
@@ -0,0 +1,126 @@
+package org.wso2.andes.subscription;
```
hastef88: License needed

akalankapagoda: yes, missed it. Will add to all new classes.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/StorageQueueData.java
akalankapagoda: Yes, Will update.

PR #498, opened 2016-01-06
[MB-1599] Coordinator node should be started first to start other nodes

sajinidesilva:
jira: https://wso2.org/jira/browse/MB-1599

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/MBThriftClient.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/MBThriftClient.java
@@ -86,7 +86,7 @@ public class MBThriftClient {
 
         } catch (ThriftClientException e) {
             handleCoordinatorChanges();
-            throw new ConnectionException(e.getMessage());
+            throw new ConnectionException("Error occurred in thrift client " + e.getMessage(), e);
         }
     }
 
@@ -132,7 +132,7 @@ public class MBThriftClient {
             }
 
         } catch (ThriftClientException e) {
-            log.error("Error occurred while receiving coordinator details from map");
+            log.error("Error occurred while receiving coordinator details from map", e);
             handleCoordinatorChanges();
         }
     }
@@ -163,7 +163,7 @@ public class MBThriftClient {
                 throw new ConnectionException("Coordinator has changed", e);
             }
         } catch (ThriftClientException e) {
-            log.error("Error occurred while receiving coordinator details from map");
+            log.error("Error occurred while receiving coordinator details from map", e);
             handleCoordinatorChanges();
         }
         return deleteSuccess;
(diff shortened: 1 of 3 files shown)


sajinidesilva:
Public jira : https://wso2.org/jira/browse/MB-1599

modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/MBThriftClient.java
```diff
@@ -212,24 +228,30 @@ public static synchronized void clearAllActiveSlotRelationsToQueue(String queueN
...
+                int thriftCoordinatorServerPort = Integer.parseInt(thriftCoordinatorServerPortString);
+                transport = new TSocket(thriftCoordinatorServerIP,
+                        thriftCoordinatorServerPort);
+                try {
+                    transport.open();
+                    TProtocol protocol = new TBinaryProtocol(transport);
+                    return new SlotManagementService.Client(protocol);
+                } catch (TTransportException e) {
+                    log.error("Could not initialize the Thrift client. " + e.getMessage(), e);
+                    throw new TTransportException(
```
ramith: this is unusual.... why do we have to wrap using a another exception?

modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/MBThriftClient.java
```diff
@@ -372,6 +394,9 @@ public static synchronized long updateSlotDeletionSafeZone(long safeZoneMessageI
                 handleCoordinatorChanges();
                 throw new ConnectionException("Coordinator has changed", e);
             }
+        } catch (ThriftClientException e) {
+            log.error("Error occurred while receiving coordinator details from map");
```
ramith: shouldn't we want to log the 'e' as well?

modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/exception/ThriftClientException.java
```diff
@@ -0,0 +1,39 @@
...
+public class ThriftClientException extends Exception {
+
+    public ThriftClientException() {
+    }
+
+    public ThriftClientException(String message) {
+        super(message);
+    }
+
+    public ThriftClientException(Throwable cause) {
```
ramith: do we have a usage for passing only a throwable without a meaningful message?

modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/MBThriftClient.java
```diff
@@ -83,6 +84,9 @@ public static synchronized Slot getSlot(String queueName,
                 throw new ConnectionException("Coordinator has changed", e);
             }
 
+        } catch (ThriftClientException e) {
+            handleCoordinatorChanges();
+            throw new ConnectionException(e.getMessage());
```
ramith: only the message? any particular reason?

modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/MBThriftClient.java
sajinidesilva: This will add more details to the general exception.

PR #501, opened 2016-01-07
[MB-1482] Fix condition to distinguish redeliveries

hastef88:
JMS transactions sometimes resulted in loss of message order. The reason was that multiple messages were being published on the same exact timestamp, and the below condition could not handle it. 

Added a new factor to the condition as a fix.

### DIFF
--- a/modules/andes-core/client/src/main/java/org/wso2/andes/client/BasicMessageConsumer.java
+++ b/modules/andes-core/client/src/main/java/org/wso2/andes/client/BasicMessageConsumer.java
@@ -1019,11 +1019,13 @@ public abstract class BasicMessageConsumer<U> extends Closeable implements Messa
                 if (o instanceof AbstractJMSMessage)
                 {
                     try {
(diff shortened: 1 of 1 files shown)


modules/andes-core/client/src/main/java/org/wso2/andes/client/BasicMessageConsumer.java
```diff
@@ -1019,7 +1019,7 @@ public void rollbackPendingMessages()
                 if (o instanceof AbstractJMSMessage)
                 {
                     try {
-                        if ((lastRollbackedMessageTimestamp > 0) && (lastRollbackedMessageTimestamp >= ((AbstractJMSMessage) o).getJMSTimestamp())) {
+                        if ((lastRollbackedMessageTimestamp > 0) && ((AbstractJMSMessage)o).getJMSRedelivered() && (lastRollbackedMessageTimestamp >= ((AbstractJMSMessage) o).getJMSTimestamp())) {
```
ramith: looks too long.

PR #511, opened 2016-01-17
Fixing https://wso2.org/jira/browse/MB-1609

ayyoob:


modules/orbit/andes/pom.xml
```diff
@@ -163,6 +163,7 @@
                         <Include-Resource>src/main/resources</Include-Resource>
                         <Bundle-Activator>org.wso2.andes.wso2.internal.QpidBundleActivator
                         </Bundle-Activator>
+                        <DynamicImport-Package>*</DynamicImport-Package>
```
hemikak: I think dynamic imports are discouraged.

ramith:
unless its really really required.
whats your package name?

modules/orbit/andes/pom.xml
ayyoob: yes it is discouraged if we are not allowing custom classes to be loaded. Since we have the configuration in broker.xml to load custom authenticator for mqtt, we should allow the class loader to able to load any classes.

ayyoob:
The package name is org.wso2.carbon.andes.authentication.mqtt

ramith:
how about we import : 
 org.wso2.carbon.andes.extensions.* (or something similar ?)
rather than putting a dynamic import?

hemikak:
+1  @ramith 's idea

ayyoob:
+1, I ll update and send

PR #513, opened 2016-01-18
[MB-1572] Removing existing durable mqtt subscriptions when the same client connects with clean session true.

hemikak:
JIRA - https://wso2.org/jira/browse/MB-1572


ramith:
@hemikak shall we re-visit this PR ? checking for AndesQueues per new connection might lead to a performance impact.

PR #499, opened 2016-01-06
[MB-1552] [AMQP] Provide support for prefixed wildcards as per specification

a5anka:
We were checking if a destination contains ".*" or ".#" to identiify
a wildcard destination in AMQPUtils#isWildCardDestination.
Removed "." character infront of wild card character to identify
wildcard prefixed destinations (such as *.SriLanka)

https://wso2.org/jira/browse/MB-1552

PR #515, opened 2016-01-22
Fixing AMQP prefixed wildcards according to spec.

akalankapagoda:
Adding changes in https://github.com/wso2/andes/pull/499 in a new pull request with minor changes.

"We were checking if a destination contains ".*" or ".#" to identiify
a wildcard destination in AMQPUtils#isWildCardDestination.
Removed "." character infront of wild card character to identify
wildcard prefixed destinations (such as *.SriLanka)

https://wso2.org/jira/browse/MB-1552"

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPUtils.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPUtils.java
@@ -49,7 +49,6 @@ import org.wso2.andes.server.store.StorableMessageMetaData;
 import org.wso2.andes.server.store.StoredMessage;
 import org.wso2.andes.server.subscription.Subscription;
 import org.wso2.andes.store.StoredAMQPMessage;
-import org.wso2.andes.amqp.AMQPLocalSubscription;
 import org.wso2.andes.subscription.LocalSubscription;
 import org.wso2.andes.subscription.OutboundSubscription;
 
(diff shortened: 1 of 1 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPUtils.java
```diff
@@ -453,13 +452,23 @@ public static boolean isTargetQueueBoundByMatchingToRoutingKey(String queueBound
         if (queueBoundRoutingKey.equals(messageRoutingKey)) {
             isMatching = true;
         } else if (queueBoundRoutingKey.indexOf(TOPIC_AND_CHILDREN_WILDCARD) > 1) {
-            String p = queueBoundRoutingKey.substring(0, queueBoundRoutingKey.indexOf(TOPIC_AND_CHILDREN_WILDCARD));
-            Pattern pattern = Pattern.compile(p + IMMEDIATE_CHILDREN_WILDCARD);
+            int wildcardIndex = queueBoundRoutingKey.indexOf(TOPIC_AND_CHILDREN_WILDCARD);
+            int partitionIndex = wildcardIndex == 0 ? 0 : wildcardIndex - 1;
```
sdkottegoda: I think the usage of 'if' clauses recommended over '?' mark.

akalankapagoda: yes, I would also prefer it that way.

PR #518, opened 2016-01-27
Fix MB-1611

hastef88:
Revert node-wise safe zone to slots that are still filling up to avoid deleting slots before completion.


ramith:
This commit needs to be cleaned up.

hastef88:
This PR has been cleaned and re-submitted through new PR : https://github.com/wso2/andes/pull/523. Therefore closing.

PR #519, opened 2016-01-29
[MB-1613] Deleting queue and non-durable topic subscriptions upon restart.

hemikak:
JIRA - https://wso2.org/jira/browse/MB-1613


akalankapagoda:
This doesn't close the subscriptions because when the subscriptions are loaded to memory from 'reloadSubscriptionsFromStorage' method, they are loaded as BasicSubscription objects. But in 'closeAllClusterSubscriptionsOfNode' method, they are passed as LocalSubscription objects and hence, the equals method in LocalSubscription class fails to detect these as the same subscriptions.

akalankapagoda:
Will reopen with changes.

PR #520, opened 2016-01-29
Fixing the connection leak in prepared statements 

pamod:
The most desired option is to enclose the connection/prepared statements between Try-with-Resources. http://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

This should be reflected in all database operations. Hence it will be tabled for future


ramith:
+1

PR #522, opened 2016-01-30
Fix for fresh slot deleted before overlapped slot is delivered

Asitha:
Additional check added for slot deletion executor to check whether there are messages in database before trying to delete a slot.

There is a scenario where a slot scheduled to delete, becoming overlapped and then assigned to a
node after some time. Then coordinator will allow to delete the slot while overlapped slot messages
are in delivery path (race condition). To avoid this a DB message count check is added before querying
coordinator for a slot deletion. This guarantees that slots are not deleted until messages in overlapped
slots are successfully delivered.

JIRA: https://wso2.org/jira/browse/MB-1612

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessageStore.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessageStore.java
@@ -134,6 +134,18 @@ public interface MessageStore extends HealthAwareStore{
                                                    long lastMsgID) throws AndesException;
 
     /**
+     * Get number of messages in the queue within the message id range
+     *
+     * @param storageQueueName name of the queue
+     * @param firstMessageId starting message id of the range
+     * @param lastMessageId end message id of the range
+     * @return number of messages for the queue within the provided message id range
+     * @throws AndesException
+     */
+    long getMessageCountForQueueInRange(final String storageQueueName, long firstMessageId, long lastMessageId)
+            throws AndesException;
+
+    /**
      * read  a metadata list from store specifying a starting message id and a count
      *
      * @param storageQueueName name of the queue
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessagingEngine.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessagingEngine.java
@@ -524,6 +524,20 @@ public class MessagingEngine {
     }
 
     /**
+     * Get number of messages in the queue within the message id range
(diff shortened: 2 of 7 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessagingEngine.java
```diff
@@ -524,6 +524,20 @@ public long getMessageCountOfQueue(String queueName) throws AndesException {
     }
 
     /**
+     * Get number of messages in the queue withing the message id range
```
ramith: withing is a word ?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessageStore.java
```diff
@@ -134,6 +134,18 @@ void updateMetadataInformation(String currentQueueName, List<AndesMessageMetadat
                                                    long lastMsgID) throws AndesException;
 
     /**
+     * Get number of messages in the queue withing the message id range
```
ramith: withing is a word ?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeletionExecutor.java
```diff
@@ -85,34 +86,37 @@ public void init() {
...
-                    //check current slot to delete is not null
-                    if (deletionAttempt != null) {
-                        //invoke coordinator to delete slot
-                        boolean deleteSuccess = deleteSlotAtCoordinator(deletionAttempt);
-                        if (!deleteSuccess) {
-                            //delete attempt not success, therefore reassign current deletion attempted slot to previous slot
-                            previouslyAttemptedSlot = deletionAttempt;
+                    // Slot to attempt current deletion
+                    Slot slot;
+                    slot = slotsToDelete.poll(1, TimeUnit.SECONDS);
```
ramith: don't this needs to be in two statements.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessagingEngine.java
Asitha: Yep, It's a word [1]. But that's not what I meant though ;) Will fix.

[1] https://www.google.com/search?client=ubuntu&channel=fs&q=withing&ie=utf-8&oe=utf-8#channel=fs&q=withing+meaning

PR #523, opened 2016-01-30
[MB-1611] and fixing an issue at the SlotDeletionExecutor

hastef88:
1. Revert node-wise safe zone to slots that are still filling up to avoid deleting slots before completion.
2. Fix SlotDeletionTask while loop and adding error logs for context.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeliveryWorker.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeliveryWorker.java
@@ -426,7 +426,7 @@ public class SlotDeliveryWorker extends Thread implements StoreHealthListener{
         if (null != slotsOfQueue) {
             slotsOfQueue.remove(slot.getId());
         } else {
-            log.warn("Slot has been deleted by the SlotDeliveryWorker before the executor can get to it. Slot : " + slot);
+            log.warn("Slot has been deleted by the SlotDeliveryWorker before the SlotDeletionExecutor can get to it. Slot : " + slot);
         }
 
     }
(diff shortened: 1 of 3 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeliveryWorker.java
```diff
@@ -421,7 +421,13 @@ public void deleteSlot(Slot slot) {
             log.debug("Releasing tracking of messages for slot " + slot.toString());
         }
         slot.deleteAllMessagesInSlot();
-        storageQueueToSlotTracker.get(slot.getStorageQueueName()).remove(slot.getId());
+
+        if (storageQueueToSlotTracker.containsKey(slot.getStorageQueueName())) {
+            storageQueueToSlotTracker.get(slot.getStorageQueueName()).remove(slot.getId());
+        } else {
```
ramith: shall we do a storageQueueToSlotTracker.get( ...) and then check for null ...
wouldn't that be faster?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotMessageCounter.java
```diff
@@ -323,11 +323,23 @@ private void updateCoordinatorWithCurrentSafezone() {
             if (slotSubmitLoopSkipCount == SLOT_SUBMIT_LOOP_SKIP_COUNT_THRESHOLD) {
                 //update current slot Deletion Safe Zone
                 try {
+
+                    long evaluatedSafeZone = currentSlotDeleteSafeZone;
+
+                    // If there are any slots pending submission to coordinator, we must lower the safe zone to their starting point.
+                    // If we do not consider pending slots at this calculation, safe zone will fly up unexpectedly.
```
ramith: what the effect of this situation (shall we explain that tooo?)

hastef88: Updated the description at jira : https://wso2.org/jira/browse/MB-1611

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeliveryWorker.java
hastef88: Fixed.

PR #526, opened 2016-01-30
Catch Throwable exceptions on critical tasks

Asitha:
Catch Throwable exceptions on critical tasks to avoid exception swallowing and eventually killing critical threads in MB.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeleteSafeZoneCalc.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeleteSafeZoneCalc.java
@@ -65,37 +65,14 @@ public class SlotDeleteSafeZoneCalc implements Runnable {
         }
         while (running) {
             if (isLive()) {
-                Set<String> nodesWithPublishedMessages;
                 try {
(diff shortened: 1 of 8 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeleteSafeZoneCalc.java
```diff
@@ -103,22 +80,50 @@ public void run() {
...
+                        Long safeZoneByPublishedMessages;
+                        try {
+                            safeZoneByPublishedMessages = SlotManagerClusterMode.getInstance()
+                                    .getLocalSafeZone(nodeID);
+
+                        } catch (AndesException e) {
+                            log.error("SlotDeleteSafeZoneCalc stopped due to failing to get last published id for node:" +
+                                    nodeID + ". Retrying after 15 seconds", e);
+                            try {
+                                Thread.sleep(TIME_TO_SLEEP_ON_ERROR);
```
ramith: we need to remove this sleep... those are bad.

Asitha: I've created a JIRA to change this implementation to a scheduled task.
Tracked by https://wso2.org/jira/browse/MB-1615

PR #528, opened 2016-01-31
[MB-1619] Fix of applying lz4

ThilankaBowala:
Fix of decompress using lz4, when the compressed message size is greater than the original message size


ramith:
We need a test case for this scenario.

PR #537, opened 2016-02-03
Delete empty slots when a member leaves the cluster

Asitha:
When a member leaves the cluster schedule to delete all the empty slots of the left node.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeletionExecutor.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeletionExecutor.java
@@ -115,7 +115,13 @@ public class SlotDeletionExecutor {
                             } else {
                                 SlotDeliveryWorker slotWorker = SlotDeliveryWorkerManager.getInstance()
                                         .getSlotWorker(slot.getStorageQueueName());
-                                slotWorker.deleteSlot(slot);
+
+                                // slotWorker can be null.
+                                // For instance if the deletion request came from coordinator after a member leaves the
+                                // cluster and no subscribers on coordinator node.
+                                if (null != slotWorker) {
+                                    slotWorker.deleteSlot(slot);
+                                }
                             }
                         } else {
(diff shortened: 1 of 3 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeliveryWorker.java
```diff
@@ -425,7 +425,8 @@ public void deleteSlot(Slot slot) {
         if (null != slotsOfQueue) {
             slotsOfQueue.remove(slot.getId());
         } else {
-            log.warn("Slot has been deleted by the SlotDeliveryWorker before the SlotDeletionExecutor can get to it. Slot : " + slot);
+            log.debug("Slot has been deleted by the SlotDeliveryWorker before the SlotDeletionExecutor can get to it." +
```
Asitha: When a subscriber leaves the node this scenario can happen. Hence this should be a debug log.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeletionExecutor.java
```diff
@@ -115,7 +115,13 @@ public void run() {
                             } else {
                                 SlotDeliveryWorker slotWorker = SlotDeliveryWorkerManager.getInstance()
                                         .getSlotWorker(slot.getStorageQueueName());
-                                slotWorker.deleteSlot(slot);
+
+                                // slotWorker can be null.
+                                // For instance if the deletion request came from coordinator after a member leaves the
+                                // cluster and no subscribers on coordinator node.
+                                if (null != slotWorker) {
+                                    slotWorker.deleteSlot(slot);
```
abeykoon: Why we have a null check here? Is there a situation slotWorker can be null?

Asitha: I have stated the scenario in the comment above. (with this fix coordinator will schedule to delete all the empty slots when a member leaves. If there are no subscribers at coordinator node, slotWorker can be null)

PR #552, opened 2016-02-16
Removed circular dependency

sajinidesilva:


modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPLocalSubscription.java
```diff
@@ -282,4 +285,9 @@ public String getStorageQueueName(String destination, String subscribedNode) {
 
         return storageQueueName;
     }
+
+    @Override
+    public String getProtocolType() {
+        return null;
+    }
```
akalankapagoda: This is giving a null pointer at the other end. Should have been verified.

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPLocalSubscription.java
```diff
@@ -72,6 +72,7 @@
     //if this subscription represent a topic subscription
     private  boolean isBoundToTopic;
 
+    private String protocolType;
```
akalankapagoda: This should be an enum, which was predefined.

modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/LocalSubscription.java
```diff
@@ -98,9 +98,9 @@ public LocalSubscription(OutboundSubscription subscription, String subscriptionI
         // In case of mock subscriptions, this becomes null
         if (null != subscription) {
             //We need to keep the protocolType in basic subscription for notification
-            if(subscription instanceof AMQPLocalSubscription) {
+            if((AMQPUtils.PROTOCOL_TYPE).equals(subscription.getProtocolType())) {
```
akalankapagoda: 1. We're doing enum vs string equals here. This is wrong.
2. Looking at the logic here, this could be simplified to one line 'setProtocolType(subscription.getProtocolType())'

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPUtils.java
```diff
@@ -71,6 +71,8 @@
 
     public static String TOPIC_EXCHANGE_NAME = "amq.topic";
 
+    public static String PROTOCOL_TYPE = "AMQP";
+
```
akalankapagoda: There is an enum defined just for this (ProtocolType), should have used it instead of defining a String here.

modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/OutboundSubscription.java
```diff
@@ -81,4 +80,10 @@ public boolean sendMessageToSubscriber(ProtocolMessage messageMetadata, AndesCon
      */
     public String getStorageQueueName(String destination, String subscribedNode);
 
+    /**
+     * Return the protocol type of the subscription.
+     * @return protocol type
+     */
```
akalankapagoda: Should have used the enum ProtocolType.

modules/andes-core/broker/src/main/java/org/wso2/andes/subscription/LocalSubscription.java
```diff
@@ -98,9 +98,9 @@ public LocalSubscription(OutboundSubscription subscription, String subscriptionI
...
                 setProtocolType(ProtocolType.AMQP);
-            } else if(subscription instanceof MQTTLocalSubscription) {
+            } else if((MQTTUtils.PROTOCOL_TYPE).equals(subscription.getProtocolType())) {
```
akalankapagoda: Logic here is to check if a string is equal to 'AMQP' or 'MQTT' and select the correct protocol type from ProtocolType.AMQP and ProtocolType.MQTT. This could have been done directly using the enums.

PR #562, opened 2016-03-08
Resolved MB-1639 - Accumulates messages in durable topics even thoughactive subscriptions subscribed to the broker

sajinidesilva:
Tracked by https://wso2.org/jira/browse/MB-1639


a5anka:
Shouldn't we merge after the code review?

PR #570, opened 2016-03-29
Fixed MB-1651

sajinidesilva:


sajinidesilva:
Public jira: https://wso2.org/jira/browse/MB-1651

PR #581, opened 2016-04-11
Fix for https://wso2.org/jira/browse/MB-1659 and introducing a separate class for client reference counting 

pamod:
In reference to https://wso2.org/jira/browse/MB-1659, Commit will fix the issue which has a thread leak in SocketConnectorIoProcessor. The PR also fixes the issue of requiring Guava dependency for client side reference counting.


ramith:
@pamod Can this be merged? We need to send this to C5 branch as well.

pamod:
@Asitha could you merge the change which adds Guvava dependency for reference counting ? Need that being merged before merging this. Also send the fix for C5 branch as well.

PR #593, opened 2016-04-19
Fix for MB-1659, adds a separate reference counter for andes client 

pamod:


pamod:
This fix was incorporated with the changes front ported by PR https://github.com/wso2/andes/pull/600

PR #597, opened 2016-04-20
[MB-1667] fix issue in durable topic message delivery with compression enabled

pumudu88:
Jira : https://wso2.org/jira/browse/MB-1667


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/disruptor/inbound/MessagePreProcessor.java
```diff
@@ -212,6 +213,16 @@ private void handleTopicRoutine(InboundEventContainer event, AndesMessage messag
                         clonedMessage.getMetadata().updateMetadata(subscription.getTargetQueue(),
                                 AMQPUtils.DIRECT_EXCHANGE_NAME);
                     }
+
+                    isCompressed = clonedMessage.getMetadata().isCompressed();
```
ramith: why do we need this variable?

pumudu88: +1 can avoid it. updated the PR.

PR #619, opened 2016-05-18
Improve ProtocolType implementation.

akalankapagoda:
Make the ProtocolType class final and pre-calculate the hashcode value and use the same in the equals method to make the operations faster.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/ProtocolType.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/ProtocolType.java
@@ -5,7 +5,7 @@ import org.apache.commons.lang.StringUtils;
 /**
  * The transport protocol type which connects with Andes core.
  */
-public class ProtocolType {
+public final class ProtocolType {
 
     /**
      * The name of the protocol.
@@ -16,6 +16,11 @@ public class ProtocolType {
      * The version of the protocol.
      */
     private String version;
+
+    /**
+     * We keep the hashCode pre calculated so that operations will be faster.
+     */
+    private int hashCode;
 
     private static String separator = "-";
 
@@ -30,6 +35,7 @@ public class ProtocolType {
     public ProtocolType(String protocolName, String version) throws AndesException {
         setProtocolName(protocolName);
         setVersion(version);
+        generateHashCode();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/ProtocolType.java
```diff
@@ -90,19 +99,30 @@ public boolean equals(Object o) {
...
-        ProtocolType that = (ProtocolType) o;
-
-        if (!protocolName.equals(that.protocolName)) return false;
-        if (!version.equals(that.version)) return false;
+        /* We're using only the pre-calculated hashcode value here since we expect to have ony a limited number
+          * of protocol types in the system and each hash code is calculated using the string values
+          * protocolName and version.
+         */
+        if (this.hashCode == o.hashCode()) {
+            return true;
```
ramith: can this be like
return o.hashCode() == this.hashCode ;

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/ProtocolType.java
```diff
@@ -90,19 +99,30 @@ public boolean equals(Object o) {
...
+    }
 
-        return true;
+    /**
+     * Since there are limited set of ProtocolType objects available, we can keep the hash code pre calculated
+     * and the the hashcode number set will only grow when a new protocol is plugged in to Andes.
+     */
+    private void generateHashCode() {
+        int result = protocolName.hashCode();
+        result = 31 * result + version.hashCode();
```
pamod: is there any reason for this to be multiplied by 31 ?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/ProtocolType.java
akalankapagoda: Yes, it can 😄 Will update.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/ProtocolType.java
akalankapagoda: Since we are using the addition of two integers to create the hash code, to distinguish between a (X + Y) and a (Y + X) we use a constant C such as hashCode = (C*X +Y). In most of the java native classes they have used the number 31 as C, I guess due to the reason mentioned in [1]. This has to do with the fact that in some old CPUs, multiplying by 31 is cheap.

[1] - http://stackoverflow.com/questions/299304/why-does-javas-hashcode-in-string-use-31-as-a-multiplier

PR #621, opened 2016-05-20
Fix MB-1678

a5anka:
- Make globalErrorBasedFlowControlEnabled variable volatile
- Throw AMQException instead of a Runtime exception when channel ID
  is not found. This done to continue network partition listener
  processing
- Process networkn partition listeners in order. A pririority level
  is introduced for each network partition listener. Additionally a
  network flag partition flag was introduced on AndesSubscription
  manager to block all subscription additions after a network
  partition was detected.


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesSubscriptionManager.java
```diff
@@ -593,6 +615,9 @@ private LocalSubscription convertClusterSubscriptionToMockLocalSubscription(Ande
      */
     @Override
     public void minimumNodeCountNotFulfilled(int currentNodeCount) {
+        synchronized (this) {
```
pumudu88: Fix for following Jira.
https://wso2.org/jira/browse/MB-1696

PR #626, opened 2016-05-30
Added initial code for resource manager feature.

hemikak:
In MB based on C4, all resources and operations on queues, topics, subscriptions and messages were exposed through an MBean JMX platform which came along with Qpid. The UI of MB was designed in a way that the JSP pages will first call a webservice through service stubs and then the webservice backend will use a JMX client which talks to the MBean server. 

Following is the architecture diagram on how it worked : 

c4-archi

So whenever we want to handle MB resources, we always have to use the Qpid JMX MBeans implementation.

For MB based on C5, the plan is to use OSGi services by having a manager class within andes core which provides services based on the type of protocol(example : amqp,mqtt) and destination type(example: queue,topic). As of now, we are moving the transport based implementation out of the andes core to support pluggable transports. Implementation for resource handlers will be at transport level as the implementations are may differ based on protocol and destination types. For example: deleting an AMQP queue is different from the how the AMQP topic is deleted. The resource handlers specific to a protocol will be registered in the manager class in andes component. 

This manager class will be exposed through OSGi by the Andes Component which can be used by the REST service or by the UI(UUF).

Following is the project structure for MB 3.5.0.

c5-mb-project-osgi-sturcture
​
Following is the proposed architecture change of exposing MB resources.

c5-mb

Class Diagram : 

c5-mb-resourcemanager 1​
​
-
Hemika​

[Architecture] [MB] Exposing resources and operations on queues, topics and subscriptions and etc with OSGi.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/server/resource/manager/AndesResourceManager.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/server/resource/manager/AndesResourceManager.java
@@ -1,0 +1,440 @@
+/*
+ * Copyright (c) 2016, WSO2 Inc. (http://wso2.com) All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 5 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/server/resource/manager/AndesResourceManager.java
```diff
@@ -0,0 +1,440 @@
...
+import org.wso2.andes.kernel.FlowControlListener;
+import org.wso2.andes.kernel.ProtocolType;
+import org.wso2.andes.server.ClusterResourceHolder;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ *
+ */
```
akalankapagoda: Need class comments.

hemikak: Fixed in https://github.com/wso2/andes/pull/627

PR #630, opened 2016-06-03
[MB-1682] Fixed issue of subscriptions being removed when nodes merge after split brain

sdkottegoda:
Addresses the issue reported at https://wso2.org/jira/browse/MB-1682

The solution for this was to modify the db sync task so that the database get first updated with the entries in the local subscriptions map, then update the cluster map according to what's in the database. This will prioritize the local subscription map over the db thus the cluster subscriptions map will be consistent with the local subscriptions map.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesContextStore.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesContextStore.java
@@ -77,9 +77,17 @@ public interface AndesContextStore extends HealthAwareStore{
      *
      * @param subscription The subscriptoin object to update
      * @throws AndesException
+     * @return If a subscription was updated return 1, else return 0
      */
-    void updateDurableSubscription(AndesSubscription subscription)
-		    throws AndesException;
+    int updateDurableSubscription(AndesSubscription subscription) throws AndesException;
+
+    /**
+     * Update already the store if there is a matching subscription. Else, insert the subscription.
+     *
+     * @param subscription The subscriptoin object to update
+     * @throws AndesException
+     */
+    void updateOrInsertDurableSubscription(AndesSubscription subscription) throws AndesException;
 
     /**
      * Updates a List of subscriptions with a given subscriptionId and given subscription information.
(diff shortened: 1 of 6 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesContextStore.java
```diff
@@ -77,9 +77,17 @@ void storeDurableSubscription(AndesSubscription subscription)
      *
      * @param subscription The subscriptoin object to update
      * @throws AndesException
+     * @return If a subscription was updated return 1, else return 0
      */
-    void updateDurableSubscription(AndesSubscription subscription)
-		    throws AndesException;
+    int updateDurableSubscription(AndesSubscription subscription) throws AndesException;
```
abeykoon: Any reason to return a int instead of a boolean?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesContextStore.java
```diff
@@ -77,9 +77,17 @@ void storeDurableSubscription(AndesSubscription subscription)
...
+
+    /**
+     * Update already the store if there is a matching subscription. Else, insert the subscription.
+     *
+     * @param subscription The subscriptoin object to update
+     * @throws AndesException
+     */
+    void updateOrInsertDurableSubscription(AndesSubscription subscription) throws AndesException;
```
abeykoon: addOrUpdate would be a better name. Can we spit the thing to two methods?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesRecoveryTask.java
```diff
@@ -44,6 +47,7 @@
 
 	AndesContextStore andesContextStore;
 	AMQPConstructStore amqpConstructStore;
+	SubscriptionEngine subscriptionEngine;
```
abeykoon: Are these public, private class variables? Let's be specific

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesContextStore.java
sdkottegoda: query.executeUpdate() directy returns an int. We don't need to do another if and convert this to a boolean.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesContextStore.java
sdkottegoda: The priority is to update. that's why I named it as updateOrInsert can rename  if it seems odd. We already have two separate methods, here, we need to insert the subscription only if it cannot be updated(If it's not there)

PR #631, opened 2016-06-03
Added message decoder interface for transport layer implementation.

hemikak:


modules/andes-core/broker/src/main/java/org/wso2/andes/server/resource/manager/MessageDecoder.java
```diff
@@ -0,0 +1,47 @@
...
+    Map<String, String> getMessageProperties(AndesMessageMetadata andesMessageMetadata) throws AndesException;
+
+    /**
+     * Gets the message content as a string.
+     *
+     * @param andesMessage The andes message object.
+     * @return Message content as a string.
+     * @throws AndesException
+     */
+    String getMessageContentAsString(AndesMessage andesMessage) throws AndesException;
```
a5anka: Can we modify the method name to "getMessageContent"? Since the method's return type is string, we don't need to mention it in method name. Isn't it?

hemikak: +1, Will do the change

PR #647, opened 2016-06-30
Reading newly aded <deployment><mode> config from broker.xml

indikasampath2000:


modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
```diff
@@ -656,7 +656,14 @@
      * subscriptions) etc.
      */
     RECOVERY_NETWORK_PARTITIONS_MINIMUM_CLUSTER_SIZE("recovery/networkPartitionsDetection/minimumClusterSize",
-                                                     "1", Integer.class);
+            "1", Integer.class),
+
+    /**
+     * Set deployment mode relevant to the setup. There are 2 predefine values for mode - 'default' and 'standalone'.
```
ramith: specifies the deployment mode for the broker node (and cluster). possible values {default, standalone}.
     default - broker node will decide to run HA ( master/slave) or fully distributed mode. Decision is taken based on node has a clustering machanism enabled or not. if node is not configured to join a cluster it will run in HA mode (refer to axis2.xml for more information). if node can join a cluster it will start in fully clustered mode.
 
 standalone - this is the simplest mode a broker can be started. node will assume datastore is not shared with another node. therefore it will not try to coordinate with other nodes (possibly non-existant) to provide HA or clustering.

PR #651, opened 2016-07-05
[OnHold] Implementing delivery delay for requeued messages.

hemikak:
Currently when a message is being rejected by a subscriber client, the client is able to delay returning the message to the server. This is done by setting the "AndesAckWaitTimeOut" system property. 

But the message cannot be delayed from the server side. I.E when a message rejected is received, the message is requeued straight away to the subscriber client.

A delay should be able to added from the server side allowing the clients not being pushed with rejected messages.

Solution is to have a configurable value in the broker.xml which applies a delay for each rejected message before being requeued.

Each subscription(LocalSubscription) will have a delayed queue[1] which will store the rejected messages with the configuration value in broker.xml. When a rejected message is received, accepting new messages will be stopped from getting delivered to the client by using the LocalSubscription#hasRoomToAcceptMessages() method. New messages will only be accepted only when these rejected messages are accepted by the client or sent to the DLC.

There will be a separate scheduled task which consists of subscriptions that has rejected messages. The scheduler will go through each of those subscriptions and deliver the messages to the outbound disruptor. 

delayeddelivery 1

[1] - https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/DelayQueue.html

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
@@ -68,6 +68,20 @@ public enum AndesConfiguration implements ConfigurationProperty {
     COORDINATION_NODE_ID("coordination/nodeID", "default", String.class),
 
     /**
+     * By default, when a message is rejected, MB will resend the message with 0ms delay.
+     * This delay value can be configured by the following value. The configuration applied for all requeued messages.
+     * Value is in milliseconds.
+     */
+    MESSAGE_REDELIVERY_DELAY("redeliveryDelay/messageRedeliveryDelay", "0", Long.class),
+
+    /**
+     * This value will only be taken into account when "redeliveryDelay/messageRedeliveryDelay" value is greater
+     * than 0ms. This time interval represents the interval between the tasks of redelivering requeued messages.
+     * Value is in milliseconds.
+     */
+    MESSAGE_REDELIVERY_EXECUTION_INTERVAL("redeliveryDelay/executionInterval", "100", Long.class),
+
+    /**
      * The IP address to which mqtt/amqp channels should be bound.
      */
     TRANSPORTS_BIND_ADDRESS("transports/bindAddress", "*", String.class),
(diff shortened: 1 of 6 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
```diff
@@ -68,6 +68,20 @@
...
+     * Value is in milliseconds.
+     */
+    MESSAGE_REDELIVERY_DELAY("redeliveryDelay/messageRedeliveryDelay", "0", Long.class),
+
+    /**
+     * This value will only be taken into account when "redeliveryDelay/messageRedeliveryDelay" value is greater
+     * than 0ms. This time interval represents the interval between the tasks of redelivering requeued messages.
+     * Value is in milliseconds.
+     */
+    MESSAGE_REDELIVERY_EXECUTION_INTERVAL("redeliveryDelay/executionInterval", "100", Long.class),
```
ramith: I suppose for this to work,
 MESSAGE_REDELIVERY_DELAY > MESSAGE_REDELIVERY_EXECUTION_INTERVAL isn't it?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/RejectedMessage.java
```diff
@@ -0,0 +1,89 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public int compareTo(Delayed rejectedMessage) {
+        if (this == rejectedMessage) {
+            return 0;
+        }
+
+        if (rejectedMessage instanceof RejectedMessage) {
+            long diff = deliveryDelay - ((RejectedMessage) rejectedMessage).deliveryDelay;
```
ramith: can't we return deliveryDelay - ((RejectedMessage) rejectedMessage).deliveryDelay;

hemikak: Yes, I missed it. Will modify.

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
hemikak: To reduce the latency which is being added by the scheduler service, the MESSAGE_REDELIVERY_EXECUTION_INTERVAL should be less than MESSAGE_REDELIVERY_DELAY. 

But it will still work even if MESSAGE_REDELIVERY_EXECUTION_INTERVAL is greater than MESSAGE_REDELIVERY_DELAY.

PR #668, opened 2016-07-29
[MB-1708] RDBMS based coordinator election algorithm

a5anka:
In MB, we have used a coordinator based approach to manage distributed messaging algorithm in the cluster. Currently Hazelcast is used to elect the coordinator. But one issue we faced with Hazelcast is, during a network segmentation (split brain), Hazelcast can elect two or more coordinators in the cluster. This affects the correctness of the distributed messaging algorithm since there are some tables in the database that should only be edited by a single node (i.e. coordinator). 

As a solution, we are implementing an RDBMS based approach to elect the coordinator node in the cluster. By doing this we can make sure that even during a network segmentation only one node will be elected as the coordinator node since the election is happening through the database.

The algorithm will use a polling mechanism to check the validity of the nodes. To make the election algorithm scalable, only the coordinator node will be checking status of all the nodes in the cluster and it will inform other nodes through database when a member is added/left. The nodes will be only checking for the status of the coordinator node. When a node detect that coordinator is invalid it will go for a election to elect a new coordinator.

#### CLUSTER_COORDINATOR_HEARTBEAT

| Column Name | DATA Type | Other Options |
|----------------|-------------|---------------|
| **ANCHOR** | INTEGER | NOT NULL |
| NODE_ID | VARCHAR(36) | NOT NULL |
| LAST_HEARTBEAT | BIGINT   | NOT NULL |
| THRIFT_HOST | VARCHAR(512)   | NOT NULL |
| THRIFT_PORT | INT   | NOT NULL |

The reason for using bigint for heartbeat instead of a time stamp is different DB vendors use different commands to insert time stamps.

#### CLUSTER_NODE_HEARTBEAT

| Column Name | DATA Type | Other Options |
|----------------|-------------|---------------|
| **NODE_ID** | VARCHAR(36) | NOT NULL |
| LAST_HEARTBEAT | BIGINT   | NOT NULL |
| IS_NEW_NODE | TINYINT   | NOT NULL |

#### MB_MEMBERSHIP

| Column Name | DATA Type | Other Options |
|----------------|-------------|---------------|
| **EVENT_ID** | BIGINT | NOT NULL AUTO_INCREMENT|
| NODE_ID | VARCHAR(512) | NOT NULL |
| CHANGE_TYPE | TINYINT(4)    | NOT NULL |
|CHANGED_MEMBER_ID | VARCHAR(512) |NOT NULL|

### Limitations

- A delay can result when coordinator is changed due to back off

https://wso2.org/jira/browse/MB-1708


ramith:
shall we squash the commits into one?

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/CoordinationConfigurableClusterAgent.java
```diff
@@ -45,22 +40,22 @@
...
  * Hazelcast based cluster agent implementation
  */
-public class HazelcastClusterAgent implements ClusterAgent {
+public class CoordinationConfigurableClusterAgent implements ClusterAgent {
 
     /**
      * Class logger
      */
-    private Log log = LogFactory.getLog(HazelcastClusterAgent.class);
+    private Log log = LogFactory.getLog(CoordinationConfigurableClusterAgent.class);
```
sdkottegoda: log should be private static final :)

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/CoordinationConfigurableClusterAgent.java
```diff
@@ -125,51 +109,39 @@ public HazelcastClusterAgent(HazelcastInstance hazelcastInstance) {
...
+    public void memberRemoved(String nodeId) throws AndesException {
+        manager.memberRemoved(nodeId);
     }
 
-    
-    public void networkPatitionMerged(){
-    	networkPartitionDetector.networkPartitionMerged();
+    public void becameCoordinator() {
+        updateCoordinatorNodeDetailMap();
+
```
sdkottegoda: Shall we remove the extra new line?

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/RDBMSCoordinationStrategy.java
```diff
@@ -0,0 +1,471 @@
...
+        private CoordinatorElectionTask() {
+            running = true;
+        }
+
+        @Override
+        public void run() {
+            while (running) {
+
+                try {
+
```
sdkottegoda: Let's remove the unnecessary new lines

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/CoordinationConfigurableClusterAgent.java
a5anka: Fixed

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/CoordinationConfigurableClusterAgent.java
a5anka: Fixed

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/RDBMSCoordinationStrategy.java
a5anka: Fixed

PR #680, opened 2016-08-15
[MB-1716] Null pointer exception is thrown when an unauthenticated client is subscribing to mqtt topic

ayyoob:
jira: https://wso2.org/jira/browse/MB-1716

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/dna/mqtt/moquette/messaging/spi/impl/ProtocolProcessor.java
+++ b/modules/andes-core/broker/src/main/java/org/dna/mqtt/moquette/messaging/spi/impl/ProtocolProcessor.java
@@ -833,11 +833,14 @@ public class ProtocolProcessor implements EventHandler<ValueEvent>, PubAckHandle
         }
 
         MQTTAuthorizationSubject authSubject = authSubjects.get(clientID);
-        if (authSubject == null) {
-            return;
-        }
         SubAckMessage ackMessage = new SubAckMessage();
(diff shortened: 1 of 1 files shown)


modules/andes-core/broker/src/main/java/org/dna/mqtt/moquette/messaging/spi/impl/ProtocolProcessor.java
```diff
@@ -833,6 +833,9 @@ void processSubscribe(ServerChannel session, SubscribeMessage msg, String client
         }
 
         MQTTAuthorizationSubject authSubject = authSubjects.get(clientID);
+        if (authSubject == null) {
```
a5anka: Can you please explain the reason for this Null check? 
Should we send a failure ack in this scenario?

ayyoob: when a client (version less than 3.1.1) is connected to the broker and when it tries to subscribe to a topic which is not authorized to, then the server will disconnect the client forcefully. However if the client does not acknowledge the message then it will try to subscribe back to the topic which would lead to a null pointer exception. 
Need to send a failure ack and have fixed it.

ramith:
Could you create a jira and add that in the PR description.

pumudu88:
@ayyoob Can you please verify in older mqtt specs (mqtt 3.1) if this behavior is defined?

PR #683, opened 2016-08-22
Introduce new subscription model

abeykoon:
Introduce new subscription model for Andes. Now it follows AMQP model with following modifications. 

1. A queue cannot bound to multiple exchanges (message routers)
2. Queue is created in persistent storage only at the binding call. This is because we create a temporary queues which is shared by local subscribers only (queue is local to the node) depending on the binding type (i.e topic, queues)
3. Introduced two new message routers for DLC messages and MQTT messages. Now MQTT has a separate message router. 

To filter subscriptions library [1] is used. In critical path it is used to lookup the subscription by channel ID. For UI and to act upon cluster split, node kill we need to query subscriptions with different criteria. Now the subscriptions can be filtered using below criteria. 

SUB_ID
PROTOCOL
NODE_ID
CHANNEL_ID  -->> at critical path
ROUTER_NAME
ROUTING_KEY
STORAGE_QUEUE_NAME
STATE

Now the subscription object is created at connection and removed at disconnection. There is no subscription to represent inactive state. Inactive subscriptions are derived from the existing durable queues bound to message routers. 

By this way Subscription model is greatly simplified.

Additionally, re-structured and re-adapted cluster coordination concepts and added RDBMS cluster coordination. Additional modifications done are as follows.

>>revamped handling of exchange changes.
>>added changes necessary for queues and subscription creation and syncing.
>>revamped subscription notifications and notification handling.
>>removed SubscriptionEngine class.
>>did necessary changes to delivery model to suit with new subscription model.
>>added changes to outbound message delivery path to suit new subscription model.
>>Adjusted inbound message received and message preprocessing and topic matching to suit new model.
>>Preventing HZ notification reaching the local node. Implement a method to get inactive subscriber representations. Modified PersistenceStoreConnector to suit new changes.
>>Updated SubscriptionManagementInformationMBean to plug new model to UI.
>>Fixed subscription ID generation.
>>removed AndesQueue and QueueListener interface.
>>Removed AndesExchange and AndesQueue classes. Responsibilities of these and moved to MessageRouter and StorageQueue classes respectively.
>>Fixed an issue identified by CleanSessionTestCase.
>>Added Inactive Subscriber Representation for durable topic subscribers.
>>Fixed issue when moving messages from DLC to a queue.
>>Added new MBean operation to get names of all durable queues.

[1]. https://github.com/npgall/cqengine


sajinidesilva:
Can we put the jira associated to this feature here?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesKernelBoot.java
```diff
@@ -264,23 +277,52 @@ private static MessageStore createMessageStoreFromConfig(AndesContextStore andes
...
+        // directly wire the instance without wrapped instance
+        messageStore = new FailureObservingMessageStore(createMessageStoreFromConfig(contextStoreInConfig));
+        // Setting the message store in the context store
+        AndesContext.getInstance().setMessageStore(messageStore);
+
+        //create AMQP Constructs store
+        amqpConstructStore = new AMQPConstructStore(contextStore);
+        AndesContext.getInstance().setAMQPConstructStore(amqpConstructStore);
+
+        //create MessageRouter Registry
```
sajinidesilva: Start comments with capital letters

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/disruptor/inbound/InboundQueueSyncEvent.java
```diff
@@ -0,0 +1,190 @@
+/*
+ *
+ *  * Copyright (c) 2015, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *  *
+ *  * WSO2 Inc. licenses this file to you under the Apache License,
```
sajinidesilva: Update copyright headers

PR #685, opened 2016-08-26
[MB-1736] Improving Thrift Based Slot Coordination for NATed/Dockerised Environments

isurulucky:
In MB's clustered deployment [1] Thrift based slot coordination requires TCP point-to-point connections. The IP and Port are read from broker.xml currently. However, in a NATed environment or in some Docker environment without an overlay network, the local IP is not visible to the other instance of MB. Therefore once the Thrift server binds to the local IP, slot coordination communication with other nodes fail. A possible solution is to bind the Thrift to the local IP + port and advertise the public/routable IP + port for the other nodes.

[1]. https://docs.wso2.com/display/CLUSTER44x/Clustering+MB+3.0.0

This PR relates to : https://github.com/wso2/carbon-business-messaging/pull/336

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
@@ -59,6 +59,20 @@ public enum AndesConfiguration implements ConfigurationProperty {
      */
     COORDINATION_THRIFT_SERVER_PORT
             ("coordination/thriftServerPort", "7611", Integer.class),
+
+    /**
+     * The IP of the Thrift server to be advertised to external. This is used in NAT/containerized
+     * environments which has a local private IP and a routable public IP.
+     */
+    COORDINATION_THRIFT_SERVER_ADVERTISED_HOST
+            ("coordination/thriftServerAdvertisedHost", "", String.class),
+
+    /**
+     * The port of the Thrift server to be advertised to external. This is used in NAT/containerized
+     * environments which has a local private IP and a routable public IP.
+     */
+    COORDINATION_THRIFT_SERVER_ADVERTISED_PORT
+            ("coordination/thriftServerAdvertisedPort", "-1", Integer.class),
             
     /**
      * Socket timeout for thrift connection.
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/CoordinationConfigurableClusterAgent.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/CoordinationConfigurableClusterAgent.java
@@ -204,13 +204,45 @@ public class CoordinationConfigurableClusterAgent implements ClusterAgent {
             log.debug("Unique ID generation for message ID generation:" + uniqueIdOfLocalMember);
         }
 
-        String thriftCoordinatorServerIP = AndesContext.getInstance().getThriftServerHost();
-        int thriftCoordinatorServerPort = AndesContext.getInstance().getThriftServerPort();
-        InetSocketAddress thriftAddress = new InetSocketAddress(thriftCoordinatorServerIP, thriftCoordinatorServerPort);
+        InetSocketAddress thriftAddress = new InetSocketAddress(getThriftCoordinatorServerAddress(), getThriftCoordinatorServerPort());
 
         coordinationStrategy.start(this, getLocalNodeIdentifier(), thriftAddress);
 
         networkPartitionDetector.start();
+    }
+
+    /**
+     * Get the advertised port of Thrift Coordinator Server
+     *
+     * @return  IP address for Thrift Coordinator Server
+     */
+    private String getThriftCoordinatorServerAddress() {
+        String thriftCoordinatorServerIP = AndesContext.getInstance().getThriftServerAdvertisedHost();
+        if (!thriftCoordinatorServerIP.isEmpty()) {
+            if (log.isDebugEnabled()) {
+                log.debug("Thrift Coordinator Server bind IP specified as " + thriftCoordinatorServerIP);
+            }
+        } else {
+            thriftCoordinatorServerIP = AndesContext.getInstance().getThriftServerHost();
+        }
+        return thriftCoordinatorServerIP;
+    }
+
+    /**
+     * Get the advertised address of Thrift Coordinator Server
+     *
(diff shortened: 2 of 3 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/CoordinationConfigurableClusterAgent.java
```diff
@@ -204,16 +204,52 @@ public void start(ClusterManager manager) throws AndesException{
...
+     * Get the routable IP address of Thrift Coordinator Server
+     *
+     * @return  IP address for Thrift Coordinator Server
+     */
+    private String getThriftCoordinatorServerAddress() {
+        // first check if the advertised routable ip is present
+        String thriftCoordinatorServerIP = AndesContext.getInstance().getThriftServerAdvertisedHost();
+        if (!thriftCoordinatorServerIP.isEmpty()) {
+            if (log.isDebugEnabled()) {
+                log.debug("Thrift Coordinator Server advertised IP specified as " + thriftCoordinatorServerIP);
```
imesh: I think if we can move this debug log statement to the line before returning the thriftCoordinatorServerIP, it would print the server ip in both cases. Thanks!

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/CoordinationConfigurableClusterAgent.java
```diff
@@ -204,16 +204,52 @@ public void start(ClusterManager manager) throws AndesException{
...
+        int thriftCoordinatorServerPort = AndesContext.getInstance().getThriftServerAdvertisedPort();
+        if (thriftCoordinatorServerPort > 0) {
+            if (log.isDebugEnabled()) {
+                log.debug("Thrift Coordinator Server advertised port specified as " + thriftCoordinatorServerPort);
+            }
+        } else {
+            // advertised ip is not defined; use the local port
+            thriftCoordinatorServerPort = AndesContext.getInstance().getThriftServerPort();
+        }
+        return thriftCoordinatorServerPort;
```
imesh: Might be better to move the debug log statement to the line number 248. Thanks!

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
```diff
@@ -59,6 +59,20 @@
      */
     COORDINATION_THRIFT_SERVER_PORT
             ("coordination/thriftServerPort", "7611", Integer.class),
+
+    /**
+     * The IP of the Thrift server to be advertised to external. This is used in NAT/containerized
+     * environments which has a local private IP and a routable public IP.
+     */
+    COORDINATION_THRIFT_SERVER_ADVERTISED_HOST
+            ("coordination/thriftServerAdvertisedHost", "", String.class),
```
imesh: Is there any specific reason to make the default hostname blank? If not, wouldn't it better to set it to localhost, as we have set the default server port to 7611? Thanks!

imesh:
Great work Isuru!! Well written!!
I think it might be useful for explaining why this feature was needed in the PR description. Thanks!

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
ramith: +1 for @imesh  change.

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
```diff
@@ -59,6 +59,20 @@
...
+
+    /**
+     * The port of the Thrift server to be advertised to external. This is used in NAT/containerized
+     * environments which has a local private IP and a routable public IP.
+     */
+    COORDINATION_THRIFT_SERVER_ADVERTISED_PORT
+            ("coordination/thriftServerAdvertisedPort", "-1", Integer.class),
```
ramith: @isurulucky @imesh I propose this also should be something other than -1. thoughts?

imesh: Yes, a good finding @ramith! It might be better to set this to a default value.

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
isurulucky: The reason is that thriftServerAdvertisedHost is optional, we do not have default values for it. Its not the IP that thrift server binds to, which is a mandatory property, but rather the advertised IP.

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
isurulucky: Again, this is not a mandatory property. If we set it to a default value, AFAIU we won't be able to distinguish if the user wants to use this property or not.

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/CoordinationConfigurableClusterAgent.java
isurulucky: +1

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/CoordinationConfigurableClusterAgent.java
isurulucky: +1

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
imesh: Yes, that make sense!

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
imesh: Yes, that make sense! Thanks!

PR #686, opened 2016-08-29
[MB-1727] - Subscriptions don't become inactive once the db connection is dropped and restored

sdkottegoda:
Addresses the issue reported at https://wso2.org/jira/browse/MB-1727

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesRecoveryTask.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesRecoveryTask.java
@@ -138,7 +138,12 @@ public class AndesRecoveryTask implements Runnable, StoreHealthListener {
 	public void storeOperational(HealthAwareStore store) {
 		if (store.getClass().getSuperclass().isInstance(AndesContextStore.class)) {
 			isContextStoreOperational.set(true);
-			log.info("AndesRecoveryTask became operational.");
-		}
+			log.info("AndesRecoveryTask became operational. Recovering broker artifacts.");
+            try {
+                recoverBrokerArtifacts();
+            } catch (AndesException e) {
+                log.error("Error in running andes recovery task", e);
+            }
+        }
 	}
 }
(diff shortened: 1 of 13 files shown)


ramith:
Could we attach a jira ID and put a descriptive name of the PR.
and make it a one commit? @sdkottegoda

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/FlowControlManager.java
```diff
@@ -376,6 +376,7 @@ public synchronized void prepareChannelsForShutdown () {
     @Override
     public void storeNonOperational(HealthAwareStore store, Exception ex) {
         //Remove all local subscriptions and enable error based flow control when the store becomes non-operational.
+        System.out.println("Flow control manager: store non operational.");
```
ramith: why are we doing a SysOUT ?

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/ClusterManager.java
```diff
@@ -314,6 +314,7 @@ public boolean getStoreHealth() {
      */
     @Override
     public void storeNonOperational(HealthAwareStore store, Exception ex) {
+        System.out.println("Cluster manager: store non operational.");
```
ramith: hope this sysOUT will not get merged.

modules/andes-core/broker/src/main/java/org/wso2/andes/store/FailureObservingAndesContextStore.java
```diff
@@ -59,12 +67,16 @@
      */
     private ScheduledFuture<?> storeHealthDetectingFuture;
 
+    private AtomicBoolean isStoreAvailable;
```
ramith: we will document this right?

pumudu88:
Can we add some description on fixed issues?

a5anka:
It is better to have a meaningful commit message

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesRecoveryTask.java
```diff
@@ -138,7 +138,12 @@ public void storeNonOperational(HealthAwareStore store, Exception ex) {
 	public void storeOperational(HealthAwareStore store) {
 		if (store.getClass().getSuperclass().isInstance(AndesContextStore.class)) {
```
sdkottegoda: Why would we need this check?

PR #709, opened 2016-09-24
Fix hard coded value for maxNumberOfMessagesToKeepInMemory

abeykoon:
Fix hard coded value for maxNumberOfMessagesToKeepInMemory. Now reading from config.


abeykoon:
Fixed issue https://wso2.org/jira/browse/MB-1793

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/SubscriberConnection.java
```diff
@@ -143,7 +147,7 @@ public long getSubscribeTime() {
      * Forcefully disconnects protocol subscriber connection from server. This is initiated by a server admin using the
      * management console.
      *
-     * @throws AndesException
+     * @throws AndesException on an protocol level issue disconnecting
```
sdkottegoda: Wrong grammer :)

abeykoon: This is fixed.

PR #717, opened 2016-09-30
[MB-1784] - Incorrect DLC name for durable subscriptions in tenants

sdkottegoda:
Addresses the issue reported at https://wso2.org/jira/browse/MB-1784

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesConstants.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesConstants.java
@@ -35,4 +35,9 @@ public class AndesConstants {
      */
     public static final String TENANT_SEPARATOR = "/";
 
+    /**
+     * The prefix which distinguishes a durable subscription.
+     */
+    public static final String DURABLE_SUBSCRIPTION_QUEUE_PREFIX = "carbon:";
+
 }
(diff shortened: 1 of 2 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesConstants.java
```diff
@@ -35,4 +35,9 @@
      */
     public static final String TENANT_SEPARATOR = "/";
 
+    /**
+     * The prefix which distinguishes a durable subscription.
+     */
+    public static final String DURABLE_SUBSCRIPTION_QUEUE_PREFIX = "carbon:";
+
```
akalankapagoda: The value 'carbon' does not suggest any durable property. Can we rename this to something like 'carbon:durable"? This will be useful when viewing these raw data from the database for debugging.

sdkottegoda: The thing is, this is what we get as the identifier for the name of the storage queue for a durable subscription. I'm just dropping it when creting the DLC queue.

Changing "carbon:" to "carbon:durable" for the storage queue requires a large amount of work since we have used this in many places. Besides it'll require having to migrate data when upgrading from previous versions. Chaging this will only improve readability. So, I think leaving it as it is, is best.

indikasampath2000: +1 to leaving as it is.

PR #718, opened 2016-10-03
[MB-1822] Fixed null pointer issue when closing subscribers.

abeykoon:
Fixing https://wso2.org/jira/browse/MB-1822.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscription.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscription.java
@@ -312,8 +312,6 @@ public class AndesSubscription {
                         + " on non durable subscription close");
             }
         }
-
-        this.subscriberConnection = null;
     }
 
     /**


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscription.java
```diff
@@ -312,8 +312,6 @@ public void closeConnection(UUID channelID, String nodeID) throws AndesException
                         + " on non durable subscription close");
             }
         }
-
-        this.subscriberConnection = null;
```
a5anka: Why is it wrong to set it to null?

abeykoon: CQEngine indexes subscriptions by connected node etc. It needs this to be not null at any time we query. 

when closing subscriptions

1. close connection -- > here is where above code made it null
2. remove from subscription registry (CQEngine access)
 
At step 2 it looks for the values. Thus keeping it until removal is not a problem as we remove the subscription everywhere  we close it.

ramith: Can we have jira ticket id in title

abeykoon: Done. Updated the title

PR #721, opened 2016-10-04
[MB-1753] Fix expiration data removal when messages reroute/restore f…

sirojan24:
Fixed the expiration data removal when the messages are rerouted/restored to queues from DLC.

Associated JIRA Link :- https://wso2.org/jira/browse/MB-1753

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesMessageMetadata.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesMessageMetadata.java
@@ -62,10 +62,6 @@ public class AndesMessageMetadata implements Comparable<AndesMessageMetadata> {
      * The timestamp at which the message arrived at the first gates of the broker.
      */
     long arrivalTime;
-    /**
-     * Whether Expiration value is defined for the message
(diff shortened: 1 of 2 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesMessageMetadata.java
```diff
@@ -284,6 +284,7 @@ public AndesMessageMetadata shallowCopy(long messageId) {
         clone.retain = retain;
         clone.metadata = metadata;
         clone.expirationTime = expirationTime;
+        clone.isExpirationDefined = isExpirationDefined;
```
akalankapagoda: Not something that has been introduced by this change, but something that can be fixed.

The variable name 'isExpirationDefined' violates best practices. The variable name should be 'expirationDefined'. The getter method for this should be 'isExpirationDefined'.

sirojan24: Fixed :)

sirojan24:
yes. this is fine

PR #722, opened 2016-10-04
[MB-1831] npe in slot deletion executor

sdkottegoda:
Addresses the issue reported at https://wso2.org/jira/browse/MB-1831

The issues was caused when slots for a storage queue are being deleted after the queue is removed.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/MessageDeliveryTask.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/MessageDeliveryTask.java
@@ -73,7 +73,7 @@ final class MessageDeliveryTask extends Task {
 
             // Get a slot from coordinator.
             Slot currentSlot = requestSlot(storageQueueName);
-
+            currentSlot.setStorageQueue(storageQueue);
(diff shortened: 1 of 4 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/MessageDeliveryTask.java
```diff
@@ -73,7 +73,7 @@ public TaskHint call() throws Exception {
 
             // Get a slot from coordinator.
             Slot currentSlot = requestSlot(storageQueueName);
-
+            currentSlot.setStorageQueue(storageQueue);
```
akalankapagoda: Shouldn't the coordinator set this and return? If we have to set the storage queue separately like this it implies that something is not right in the slot coordinator when returning a slot since we pass the storage queue when requesting a slot.

sdkottegoda: The thing is we only pass the 'storageQueueName' when requesting for a slot. But we need to set the storage queue object itself.

PR #724, opened 2016-10-06
[MB-1814] Fixed issue with filtering subscriptions to close when a member leaves in cluster.

abeykoon:
Fixing: https://wso2.org/jira/browse/MB-1814

This fix may affect failover scenarios as well. When one member leaves the cluster coordinator node will go and clean up subscriptions made to that node. Only local subscriptions can be closed.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscriptionManager.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscriptionManager.java
@@ -607,14 +607,17 @@ public class AndesSubscriptionManager implements NetworkPartitionListener, Store
         }
     }
 
-    public AndesSubscription getSubscriptionByNode(String nodeID, ProtocolType
-            protocolType) {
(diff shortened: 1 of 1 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscriptionManager.java
```diff
@@ -601,15 +601,18 @@ public AndesSubscription getSubscriptionById(String  subscriptionId) {
...
 
-    public AndesSubscription getSubscriptionByNode(String nodeID, ProtocolType
-            protocolType) {
-        Query<AndesSubscription> subscriptionQuery = and
-                (equal(AndesSubscription
-                        .NODE_ID, nodeID), equal(AndesSubscription
-                        .PROTOCOL, protocolType));
-        return subscriptionRegistry.exucuteQuery(subscriptionQuery).iterator().next();
+    /**
+     * Get all subscriptions connected to given node
```
sdkottegoda: Need to end with a period

PR #727, opened 2016-10-10
[MB-1723]  Fix Node ID is not set into Hazelcast map 

abeykoon:
Fix is trying to read nodeID in an exponentially back-off way.
Fixed: https://wso2.org/jira/browse/MB-1723

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/CoordinationConfigurableClusterAgent.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/CoordinationConfigurableClusterAgent.java
@@ -136,18 +136,39 @@ public class CoordinationConfigurableClusterAgent implements ClusterAgent {
     public void becameCoordinator() {
         manager.localNodeElectedAsCoordinator();
     }
-    
+
     /**
-     * Get id of the give node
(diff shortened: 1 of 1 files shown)


sdkottegoda:
My understanding was that we'll add this logic when a member is added, before logging the member name. Adding the logic in 'CoordinationConfigurableClusterAgent' is not suitable since there are many usages of it where returning null could result in an errorneous situation.

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/CoordinationConfigurableClusterAgent.java
```diff
@@ -136,18 +136,39 @@ public void memberRemoved(String nodeId) throws AndesException {
...
+                }
+            } else {
+                return nodeId;
+            }
+
+        }
+        if (StringUtils.isEmpty(nodeId)) {
+            log.warn("Node ID is not set for node " + node + " when new member joined");
         }
         return nodeId;
```
a5anka: We should return null if we cannot read the value from nodeIDMap according to the contract we agreed for network based services.

abeykoon:
Please ignore this PR.

PR #728, opened 2016-10-10
[MB-1723] Fix Node ID is not set into Hazelcast map 

abeykoon:
Fix is trying to read nodeID in an exponentially back-off way.
Fixed: https://wso2.org/jira/browse/MB-1723

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/HazelcastCoordinationStrategy.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/HazelcastCoordinationStrategy.java
@@ -21,6 +21,7 @@ import com.hazelcast.core.Member;
 import com.hazelcast.core.MemberAttributeEvent;
 import com.hazelcast.core.MembershipEvent;
 import com.hazelcast.core.MembershipListener;
+import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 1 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/HazelcastCoordinationStrategy.java
```diff
@@ -193,7 +195,35 @@ public void memberAdded(MembershipEvent membershipEvent) {
...
+                try {
+                    // Exponentially increase waiting time
+                    long sleepTime = Math.round(Math.pow(2, (numberOfAttemptsTried)));
+                    TimeUnit.SECONDS.sleep(sleepTime);
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                }
+                nodeId = configurableClusterAgent.getIdOfNode(member);
+                numberOfAttemptsTried = numberOfAttemptsTried + 1;
+                if (!(StringUtils.isEmpty(nodeId))){
```
sdkottegoda: We don't need this break anymore, right?

abeykoon: Fixed with 74e123ee0b297157465d05d9cdf1d7700b4ab663

PR #729, opened 2016-10-12
[MB-1846] Optimize mb slot recovery process during startup

pumudu88:
Fetch result set from the database once and iterate while recovering slots eliminating multiple database calls.
jira : https://wso2.org/jira/browse/MB-1846

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
@@ -35,6 +35,7 @@ import org.wso2.andes.kernel.DeliverableAndesMetadata;
 import org.wso2.andes.kernel.DurableStoreConnection;
 import org.wso2.andes.kernel.MessageStore;
 import org.wso2.andes.kernel.slot.Slot;
+import org.wso2.andes.kernel.slot.RecoverySlotCreator;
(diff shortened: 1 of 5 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSMessageStoreImpl.java
```diff
@@ -925,16 +941,20 @@ public long getMessageCountForQueueInRange(final String storageQueueName, long f
...
         PreparedStatement preparedStatement = null;
         ResultSet results = null;
+        int restoreMessagesCounter = 0;
 
-        Context nextMessageIdsRetrievalContext = MetricManager
-                .timer(Level.INFO, MetricsConstants.GET_NEXT_MESSAGE_IDS_FROM_QUEUE).start();
-        Context contextRead = MetricManager.timer(MetricsConstants.DB_READ, Level.INFO).start();
+        Context nextMessageIdsRetrievalContext = MetricManager.timer(Level.INFO,
+                MetricsConstants.GET_NEXT_MESSAGE_IDS_FROM_QUEUE)
+                .start();
```
a5anka: Can we use the same formatting we used for this line of code.

```java
Context nextMessageIdsRetrievalContext = MetricManager
        .timer(Level.INFO, MetricsConstants.GET_NEXT_MESSAGE_IDS_FROM_QUEUE).start();
```

PR #731, opened 2016-10-14
[MB-1812][ESB + MB] Fixing nullpointer when calling through JMS transport

abeykoon:
Fixing [MB-1812.](https://wso2.org/jira/browse/MB-1812) Parent qpid code also has null checks before calling dispatcher. Fixing with same.

### DIFF
--- a/modules/andes-core/client/src/main/java/org/wso2/andes/client/BasicMessageConsumer.java
+++ b/modules/andes-core/client/src/main/java/org/wso2/andes/client/BasicMessageConsumer.java
@@ -619,7 +619,9 @@ public abstract class BasicMessageConsumer<U> extends Closeable implements Messa
                         // no point otherwise as the connection will be gone
                         if (!_session.isClosed() || _session.isClosing())
                         {
-                            _session._dispatcher.rejectPending(this);
+                            if(null != _session._dispatcher) {
+                                _session._dispatcher.rejectPending(this);
+                            }
                             sendCancel();
                             cleanupQueue();
                         }


modules/andes-core/client/src/main/java/org/wso2/andes/client/BasicMessageConsumer.java
```diff
@@ -619,7 +619,9 @@ public void close(boolean sendClose) throws JMSException
                         // no point otherwise as the connection will be gone
                         if (!_session.isClosed() || _session.isClosing())
                         {
-                            _session._dispatcher.rejectPending(this);
+                            if(null != _session._dispatcher) {
```
abeykoon: See https://github.com/wso2/andes/blob/master/modules/andes-core/client/src/main/java/org/wso2/andes/client/AMQSession.java#L2581

We are on same call path.

a5anka: Can it lead to any errors if we do not `rejectPending` since the dispatcher is null?

a5anka: @akalankapagoda do you know the reason for rejecting messages here?

akalankapagoda: The reject call was brought here from AMQSession since we wanted to send reject call before the session close otherwise a NullPointer will occur in the server side.

https://github.com/wso2/andes/commit/3f39004c96980ad2578f0f5faacda59eb587f6f2

abeykoon: By underlying Apache Qpid design _session._dispatcher can be null. This is because dispatcher is only started if one of following methods are called.
consumer.receive()
consumer.receiveNoWait()
consumer.setMessageListener()
Here in axis2 transport we create the message consumer to get the queue created. But during close this is thrown as here dispatcher is not set. null check is reasonable according to design as every session does not have a dispatcher.

PR #734, opened 2016-10-17
[MB-1770] stop thrift server when shutdown call is received

a5anka:
jira: https://wso2.org/jira/browse/MB-1770

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/MBThriftServer.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/MBThriftServer.java
@@ -30,12 +30,18 @@ import org.wso2.andes.kernel.AndesException;
 import org.wso2.andes.thrift.slot.gen.SlotManagementService;
 
 import java.net.InetSocketAddress;
+import java.util.concurrent.TimeUnit;
 
 /**
- * This class is take cares of starting and stopping the thrift server which is used to do slot
- * communication.
+ * This class takes care of starting and stopping the thrift server which is used to do slot communication.
  */
 public class MBThriftServer {
+
+    /**
+     * Default thrift server request timeout. Server stop() will also wait this amount of time at max before
+     * returning. The time unit is seconds.
+     */
+    private static final int THRIFT_SERVER_REQUEST_TIMEOUT = 20;
 
     private static Log log = LogFactory.getLog(MBThriftServer.class);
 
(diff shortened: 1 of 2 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/MBThriftServer.java
```diff
@@ -30,13 +30,20 @@
 import org.wso2.andes.thrift.slot.gen.SlotManagementService;
 
 import java.net.InetSocketAddress;
+import java.util.concurrent.TimeUnit;
 
 /**
  * This class is take cares of starting and stopping the thrift server which is used to do slot
  * communication.
  */
 public class MBThriftServer {
```
Asitha: Shall we correct this to "This class ~~is~~ take care~~s~~ of..."

a5anka: Done

modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/MBThriftServer.java
```diff
@@ -30,13 +30,19 @@
...
- * communication.
+ * This class takes care of starting and stopping the thrift server which is used to do slot communication.
  */
 public class MBThriftServer {
 
+    /**
+     * Default thrift server request timeout. Server stop() will also wait this amount of time at max before
+     * returning. The time unit is seconds.
+     */
+    private static final int THRIFT_SERVER_REQUEST_TIMEOUT = 20;
```
pumudu88: Will making this a broker configuration useful? WDYT?

a5anka: We can make it a broker config if we want. But I don't think it will be very  useful for a user.

modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/MBThriftServer.java
```diff
@@ -92,12 +101,49 @@ public void start(final String hostName,
...
+        final long timeoutInMillis = unit.toMillis(timeout);
+        final long shutDownStartTime = System.currentTimeMillis();
+        long timeElapsed = 0;
+
+        while (server.isServing() && timeElapsed < timeoutInMillis) {
+            TimeUnit.MILLISECONDS.sleep(100);
+            timeElapsed = System.currentTimeMillis() - shutDownStartTime;
+        }
+
+        return timeElapsed < timeoutInMillis;
```
sajinidesilva: Isn't there a possibility where server is stopped and timeElapsed > timeoutInMillis. In that case we are returning false here and a warning will be printed. SHouldn't we return server.isServing?

a5anka:
Thank you @bsenduran. Added a description to the commit message.

modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/MBThriftServer.java
a5anka: Done

PR #735, opened 2016-10-19
Removing MQTT transport

pamod:
Removes all the dependencies related to MQTT


akalankapagoda:
😢

pamod:
This will not be implied. Since MQTT would be supported by MB

PR #739, opened 2016-10-24
[MB-1850] messages stored after durable subscription unsubscribes

sdkottegoda:
Fixes the issue reported at https://wso2.org/jira/browse/MB-1850


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/router/TopicMessageRouter.java
```diff
@@ -71,7 +71,7 @@ public TopicMessageRouter(String encodedRouterInfo) {
      * @throws AndesException
      */
     public void removeMapping(String bindingKey, StorageQueue queue) throws AndesException {
-        if (queue.getBoundSubscriptions().isEmpty() && !queue.isDurable()) {
```
a5anka: Did you remove this condition expecting non-durable will also get executed? I think with the fixes done in https://github.com/wso2/andes/pull/743 remove binding calls for non-durable queues are not reaching here.

PR #743, opened 2016-11-02
[MB-1860]Queue delete and purge in non durable path

akalankapagoda:
Fixing https://wso2.org/jira/browse/MB-1860

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
@@ -513,7 +513,7 @@ public class QpidAndesBridge {
             Andes.getInstance().addBinding(binding);
         } catch (AndesException e) {
             log.error("error while creating binding", e);
-            throw new AMQInternalException("error while removing queue", e);
+            throw new AMQInternalException("error while creating binding", e);
         }
     }
 
@@ -533,13 +533,12 @@ public class QpidAndesBridge {
                 log.debug("Ignored binding for default exchange " + exchange.getNameShortString());
             }
             return;
-        } else if (exchange.getNameShortString().equals(ExchangeDefaults.TOPIC_EXCHANGE_NAME) && !binding.getQueue()
-                                                                                                       .isDurable()) {
+        } else if (!binding.getQueue().isDurable() &&
+                    exchange.getNameShortString().equals(ExchangeDefaults.TOPIC_EXCHANGE_NAME)) {
+
             if (log.isDebugEnabled()) {
                 log.debug("Ignored binding for non durable topic " + binding.getBindingKey());
             }
-
-            log.error("Ignored binding for non durable topic " + binding.getBindingKey());
 
             return;
         }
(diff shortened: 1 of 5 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
```diff
@@ -514,7 +513,7 @@ public static void createBinding(Exchange exchange, AMQShortString routingKey,
             Andes.getInstance().addBinding(binding);
         } catch (AndesException e) {
             log.error("error while creating binding", e);
```
lankavitharana: shouldn't we add e.getMessage() to the log as well?

akalankapagoda: Logger will print the exception message when exception object is passed in. If we add e.getMessage() to the message it will be redundant and hence is not a good practice.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscriptionManager.java
```diff
@@ -262,6 +266,15 @@ private void removeLocalSubscriptionAndNotify(AndesSubscription subscription) th
         clusterNotificationAgent.notifySubscriptionsChange(subscription,
                 ClusterNotificationListener.SubscriptionChange.Closed);
 
+        // If there are no subscriptions for this queue, then delete it
+        if (storageQueue.getBoundSubscriptions().isEmpty() ) {
+
+            AndesContextInformationManager contextInformationManager = AndesContext.getInstance()
+                    .getAndesContextInformationManager();
+
+            contextInformationManager.deleteQueue(storageQueue);
```
a5anka: We should only delete storage queue for non-durable subscribers. Don't we need to check that here before deleting?

akalankapagoda: Yup, thanks. Will update the code.

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
```diff
@@ -534,6 +533,13 @@ public static void removeBinding(Binding binding) throws AndesException {
                 log.debug("Ignored binding for default exchange " + exchange.getNameShortString());
             }
             return;
+        } else if (exchange.getNameShortString().equals(ExchangeDefaults.TOPIC_EXCHANGE_NAME) && !binding.getQueue()
```
akalankapagoda: Yes, will flip. Thanks.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscriptionManager.java
akalankapagoda: Fixed.

akalankapagoda: Fixed.

PR #757, opened 2016-12-16
Fix compilation error due to unmappable character for encoding ASCII

a5anka:


a5anka:
Already fixed in https://github.com/wso2/andes/pull/758

PR #772, opened 2017-01-17
Fix multiple durable subscribers with same subscription Id test failure

Asitha:
JIRA: https://wso2.org/jira/browse/MB-1876


modules/andes-core/client/src/main/java/org/wso2/andes/client/AMQSession.java
```diff
@@ -123,6 +123,8 @@
  */
 public abstract class AMQSession<C extends BasicMessageConsumer, P extends BasicMessageProducer> extends Closeable implements Session, QueueSession, TopicSession
 {
+
+
```
hemikak: Do we need these empty lines ?

Asitha: Nope. Removing... :+1:

PR #777, opened 2017-01-20
Fixing conf directory path to be taken from system property

himasha:


modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/AndesConfigurationManager.java
```diff
@@ -110,8 +111,10 @@
      * Main file path of configuration files into which all other andes-specific config files (if any)
      * should be linked.
      */
-    private static final String ROOT_CONFIG_FILE_PATH = System.getProperty(ServerConstants.CARBON_HOME) +
-            "/repository/conf/";
+    public static final String CARBON_CONFIG_DIR_PATH = "carbon.config.dir.path";
+    public static final String CARBON_HOME = "carbon.home";
+    private static  String ROOT_CONFIG_FILE_PATH;
```
hastef88: Can make ROOT_CONFIG_FILE_PATH a non-static method level variable (given that its now a dynamic value.).

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/AndesConfigurationManager.java
```diff
@@ -145,7 +148,16 @@
      */
     public static void initialize(int portOffset) throws AndesException {
 
-        String brokerConfigFilePath = ROOT_CONFIG_FILE_PATH + ROOT_CONFIG_FILE_NAME;
+       // If system property is available set conf path to that value
+        if(componentsPath != null){
+            ROOT_CONFIG_FILE_PATH = Paths.get(componentsPath).toString();
+        }
+        else{
```
akalankapagoda: Needs to format this.

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/AndesConfigurationManager.java
himasha: Fixed with 	79139c6

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/AndesConfigurationManager.java
himasha: 79139c6

PR #789, opened 2017-02-01
NPE in the mix subscriptions test case is fixed.

indikasampath2000:
Last nondurable subscriber disconnect causes purge messages and deletes queue. In queue
delete, we again invoke close subscription and send a notification. It was repeated queue
delete again and end up with NPE. Therefore, invoking subscription close inside queue delete
was removed in the code. Also, when syncing subscription, cqengine query trying to
access a null object. Null check added to ROUTER_NAME and ROUTING_KEY.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscription.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscription.java
@@ -15,6 +15,10 @@
 
 package org.wso2.andes.kernel.subscription;
 
+import com.googlecode.cqengine.attribute.Attribute;
+import com.googlecode.cqengine.attribute.SimpleAttribute;
+import com.googlecode.cqengine.attribute.SimpleNullableAttribute;
+import com.googlecode.cqengine.query.option.QueryOptions;
 import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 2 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscription.java
```diff
@@ -425,20 +426,27 @@ public UUID getValue(AndesSubscription sub, QueryOptions queryOptions) {
...
 
-    public static final Attribute<AndesSubscription, String> ROUTER_NAME = new SimpleAttribute<AndesSubscription,
-            String>("routerName") {
+    public static final Attribute<AndesSubscription, String> ROUTER_NAME =
+            new SimpleNullableAttribute<AndesSubscription, String>("routerName") {
         @Override
         public String getValue(AndesSubscription sub, QueryOptions queryOptions) {
-            return sub.storageQueue.getMessageRouter().getName();
+            String routerName = null;
+            if (sub.storageQueue.getMessageRouter() != null) {
```
sdkottegoda: Can the message router of a storage queue be null?

indikasampath2000: Yes. This is getting null in subscription sync. It calls SubscriptionRegistry#removeSubscription. This removes AndesSubscription from cqengine IndexedCollection. Remove method propagate to cqengine query method and trying to access the null object. Before doing subscription sync, we do message router sync, queue sync, and binding sync and these are setting the object reference to null.

PR #802, opened 2017-02-09
Add config broker/transaction/maxParallelDtxChannels

a5anka:


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/Andes.java
```diff
@@ -804,8 +817,35 @@ public void triggerRecoveryEvent() {
...
-        return new DistributedTransaction(dtxRegistry, inboundEventManager, channel);
+    /**
+     * Allocate a distributed transaction object for dtx calls
+     *
+     * @param channel   channel object of the session
+     * @param sessionID session id
+     * @return created DistributeTransaction object
+     * @throws AndesException if maximum number of parallel transactions limit reached
+     */
+    public DistributedTransaction createDistributedTransaction(AndesChannel channel, long sessionID)
```
akalankapagoda: Do we need to synchronize this and the 'releaseDistributedTransaction methods? If multiple threads access this, there can be erroneous situations.

a5anka: yes. We need to synchronize

a5anka: Fixed. Accessing methods are now synchronized.

PR #804, opened 2017-02-09
MB-1815 issue fixed.

indikasampath2000:
Non durable topic subscriber connection closed through subscription list page. The subscriber connection closed without issue. But topic couldn't delete in the topic list page due to registry resource not getting cleared. The issue was user subject not passed to the authorization when closing the connection. This was fixed by changing the close connection and delete registry entry as two different calls. 

Public JIRA - https://wso2.org/jira/browse/MB-1815

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/mqtt/MQTTLocalSubscription.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/mqtt/MQTTLocalSubscription.java
@@ -235,6 +235,11 @@ public class MQTTLocalSubscription implements OutboundSubscription {
         return 0;
     }
 
+    @Override
+    public String getProtocolQueueName() {
+        return wildcardDestination;
+    }
+
     //TODO: decide how to call this
     public void ackReceived(long messageID) {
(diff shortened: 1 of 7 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/mqtt/MQTTLocalSubscription.java
```diff
@@ -235,6 +235,11 @@ public long getSubscribeTime() {
         return 0;
     }
 
+    @Override
+    public String getProtocolQueueName() {
+        return wildcardDestination;
```
a5anka: Can this lead to the registry resource getting deleted when we have multiple subscribers with the same wildcard destination? Since all the non-durable topic subscribers have the same protocol queue name in above situation, when only one subscription disconnects, we will be deleting the registry resource for other active subscriptions.

indikasampath2000: This is implemented just to return the MQTT topic name. Anyway, protocolQueue name only used in AMQP non durable topic subscription details.

PR #809, opened 2017-02-14
Fix for MB-1880

indikasampath2000:
Enable rdbmsBasedCoordination and RDBMSBasedClusterEventSynchronization caused thrift exception at server startup when electing coordinator. The problem is we are starting thrift server regardless of coordinator election. This fix is to wait until coordinator election completed and then start thrift server.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/RDBMSCoordinationStrategy.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/RDBMSCoordinationStrategy.java
@@ -212,6 +212,24 @@ class RDBMSCoordinationStrategy implements CoordinationStrategy, RDBMSMembership
 
         coordinatorElectionTask = new CoordinatorElectionTask();
         threadExecutor.execute(coordinatorElectionTask);
+
+        // Wait until node state become Candidate/Coordinator because thrift server needs to start after that.
+        int timeout = 500;
+        int waitTime = 0;
+        int maxWaitTime = heartbeatMaxAge * 5;
+        while (currentNodeState == NodeState.ELECTION) {
+            try {
+                TimeUnit.MILLISECONDS.sleep(timeout);
+                waitTime = waitTime + timeout;
+                if (waitTime == maxWaitTime) {
+                    throw new RuntimeException("Node is stuck in the ELECTION state for "
+                            + waitTime + " milliseconds.");
+                }
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new RuntimeException("An error occurred while waiting to get current node state.", e);
+            }
+        }
     }
 
     /**


indikasampath2000:
Public JIRA - https://wso2.org/jira/browse/MB-1880

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/RDBMSCoordinationStrategy.java
```diff
@@ -212,6 +212,23 @@ public void start(CoordinationConfigurableClusterAgent configurableClusterAgent,
...
         coordinatorElectionTask = new CoordinatorElectionTask();
         threadExecutor.execute(coordinatorElectionTask);
+
+        // Wait until node state become Candidate/Coordinator because thrift server needs to start after that.
+        int waitTime = 0;
+        int maxWaitTime = heartbeatMaxAge * 5;
+        while (currentNodeState == NodeState.ELECTION) {
+            try {
+                TimeUnit.MILLISECONDS.sleep(500);
+                waitTime = waitTime + 500;
```
a5anka: Better to define a variable for 500 since we have used it in a couple of places.

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/RDBMSCoordinationStrategy.java
```diff
@@ -212,6 +212,23 @@ public void start(CoordinationConfigurableClusterAgent configurableClusterAgent,
...
+                if (waitTime == maxWaitTime) {
+                    throw new RuntimeException("Node status didn't change to Candidate/Coordinator within "
+                            + waitTime + " milliseconds.");
+                }
+            } catch (InterruptedException e) {
+                logger.error("An error occurred while waiting to get current node state.", e);
+                Thread.currentThread().interrupt();
+            }
+        }
```
a5anka: How about moving this whole thing to a private method to improve readability?

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/RDBMSCoordinationStrategy.java
```diff
@@ -212,6 +212,23 @@ public void start(CoordinationConfigurableClusterAgent configurableClusterAgent,
...
+        // Wait until node state become Candidate/Coordinator because thrift server needs to start after that.
+        int waitTime = 0;
+        int maxWaitTime = heartbeatMaxAge * 5;
+        while (currentNodeState == NodeState.ELECTION) {
+            try {
+                TimeUnit.MILLISECONDS.sleep(500);
+                waitTime = waitTime + 500;
+                if (waitTime == maxWaitTime) {
+                    throw new RuntimeException("Node status didn't change to Candidate/Coordinator within "
+                            + waitTime + " milliseconds.");
```
a5anka: Maybe we can say something similar to "Node is stuck in the election for ___ milliseconds". In the future, we might add some more states other than candidate and coordinator

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/RDBMSCoordinationStrategy.java
```diff
@@ -212,6 +212,23 @@ public void start(CoordinationConfigurableClusterAgent configurableClusterAgent,
...
+                }
+            } catch (InterruptedException e) {
```
a5anka: In this scenario, if interrupted I think we should throw an exception since most probably someone pressed Ctrl-c

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/RDBMSCoordinationStrategy.java
indikasampath2000: Changes made in the e8cd302

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/RDBMSCoordinationStrategy.java
indikasampath2000: Changes made in the e8cd302

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/RDBMSCoordinationStrategy.java
indikasampath2000: Changes made in the e8cd302

modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/RDBMSCoordinationStrategy.java
indikasampath2000: As per above comment, code changed to throw RuntimeException if InterruptedException occurred.  I feel that it is not necessary to move it to private method since there is no code duplication.

PR #811, opened 2017-02-15
[MB-1703] added new message status NO_MATCHING_CONSUMER

abeykoon:
Scenario: When there is no subscriber with a matching selector if it is a queue message we place in on DLC. 

BUFFERED >> SCHEDULED_TO_SEND >> DLC (invalid transition)

With the fix

BUFFERED >> NO_MATCHING_CONSUMER >> DLC

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/server/handler/BasicRejectMethodHandler.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/server/handler/BasicRejectMethodHandler.java
@@ -127,7 +127,6 @@ public class BasicRejectMethodHandler implements StateAwareMethodListener<BasicR
             }
             else
             {
-                _logger.warn("Dropping message as requeue not required and there is no dead letter queue");
                 try {
                     DeliverableAndesMetadata andesMetadata = AndesUtils.lookupDeliveredMessage(message.getMessage()
                             .getMessageNumber(), channel.getId());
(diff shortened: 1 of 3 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/server/handler/BasicRejectMethodHandler.java
```diff
@@ -127,15 +127,15 @@ public void methodReceived(AMQStateManager stateManager, BasicRejectBody body, i
...
-                _logger.warn("Dropping message as requeue not required and there is no dead letter queue");
                 try {
                     DeliverableAndesMetadata andesMetadata = AndesUtils.lookupDeliveredMessage(message.getMessage()
                             .getMessageNumber(), channel.getId());
                     Andes.getInstance().moveMessageToDeadLetterChannel(andesMetadata, message
                             .getQueue()
                             .getName());
                 } catch (AndesException e) {
-                    _logger.error("Error while moving message to DLC" , e);
+                    _logger.error("Error while moving message to DLC message Id = "
```
lankavitharana: why are we using variable names starting with "_"? is it a standard? I feel it will be great if we can follow a single standard.

abeykoon: Understood. But the forked Qpid code has followed this convention for all the variables in all places. Thus will be keeping it.

PR #813, opened 2017-02-15
[MB-1865] Added modifications to execute slot deleting tasks in parallel.

abeykoon:
WSO2 MB reads slots, deliver messages and remove them from memory if all messages are ACKED of that slot. This deletion happens with a cluster-wide agreement (safe zone value calculation).
We have seen issues slots are getting accumulated in long running due to various reasons.
1. A few messages are not always ACKED. So slot remains
2. Slot deletion task gets inactive? 
3. Safe zone value calculation is jumbled
nodes are not time synced
heat beat stops
So it is better to know if slots are getting accumulated ahead before situation get worse. Thus need a WARN level log that prints every now and again.
Also need to make several slot deleting tasks to execute slot delete tasks in parallel (using thrift connection pool) because slot read happens in parallel (slot delivery worker threads > 1) and delivery happens in parallel. Acknowledge and slot removal should also happen in parallel to stop message accumulation in long run.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
@@ -434,7 +434,22 @@ public enum AndesConfiguration implements ConfigurationProperty {
     PERFORMANCE_TUNING_SUBMIT_SLOT_TIMER_PERIOD (
             "performanceTuning/slots/timerPeriod", "3000", Integer.class, PERFORMANCE_TUNING_SUBMIT_SLOT_TIMEOUT),
 
-    
+    /**
+     * Number of parallel threads to execute slot deletion task. Increasing this value will remove slots
+     * whose messages are read, delivered to consumers, acknowledged faster reducing heap memory used by server.
+     */
+    PERFORMANCE_TUNING_SLOT_DELETE_TASK_COUNT (
+            "performanceTuning/slots/deleteTaskCount", "5",  Integer.class),
+
+    /**
+     * Max number of pending message count to delete per Slot Deleting Task. This config is used to raise a WARN
+     * when pending scheduled number of slots exceeds this limit (indicate of an  issue that can lead to
+     * message accumulation on server.
+     */
+    PERFORMANCE_TUNING_MAX_PENDING_SLOTS_TO_DELETE (
+             "performanceTuning/slots/SlotDeleteQueueDepthWarningThreshold", "1000", Integer.class),
+
+
     /**
      * Maximum number of undelivered messages that can be in memory. Increasing this value could cause out of memory
      * scenarios, but performance will be improved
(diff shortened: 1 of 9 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeletingTask.java
```diff
@@ -0,0 +1,145 @@
...
+     * Condition running the task.
+     */
+    private boolean isLive = true;
+
+    /**
+     * Queue to keep slots to be removed. Reason to use a queue is
+     * if a slot could not be removed, it is definite later slots cannot be removed
+     * as well (because of safe-zone)
+     */
+    private LinkedBlockingQueue<Slot> slotsToDelete;
```
hastef88: when we have both slow and fast consumers for 2 queues, shouldn't we try to delete the next slot instead of waiting until the earliest slot clears up ?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeletingTask.java
```diff
@@ -0,0 +1,145 @@
...
+ * coordinate with slot coordinator in cluster to delete slots.
+ */
+public class SlotDeletingTask implements Runnable {
+
+    private static Log log = LogFactory.getLog(SlotDeletingTask.class);
+
+    /**
+     * Condition running the task.
+     */
+    private boolean isLive = true;
```
Asitha: Shall we make this volatile? Accessed by multiple threads

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
```diff
@@ -434,7 +434,22 @@
...
+    PERFORMANCE_TUNING_SLOT_DELETE_TASK_COUNT (
+            "performanceTuning/slots/deleteTaskCount", "5",  Integer.class),
+
+    /**
+     * Max number of pending message count to delete per Slot Deleting Task. This config is used to raise a WARN
+     * when pending scheduled number of slots exceeds this limit (indicate of an  issue that can lead to
+     * message accumulation on server.
+     */
+    PERFORMANCE_TUNING_MAX_PENDING_SLOTS_TO_DELETE (
+             "performanceTuning/slots/maxNumberOfPendingSlotsToDelete", "1000", Integer.class),
```
Asitha: Maybe ```PendingToDeleteSlotsWarningThreshold```? Some name from the perspective of the user not from developer.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeletingTask.java
abeykoon: If there is a slow consumer, slot will not be scheduled to be removed. This list only has slots which has no messages in range. Can we safely assume there will not be a lot of overlapped messages, which can slow down slot deletion?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotDeletingTask.java
abeykoon: Fixed.

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
abeykoon: Changed naming to SlotDeleteQueueDepthWarningThreshold

PR #817, opened 2017-02-20
Fix slot delete in standalone mode.

abeykoon:
Slots are simulated in standalone mode. Thus they need to be removed in standalone mode as well.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
@@ -404,7 +404,7 @@ public enum AndesConfiguration implements ConfigurationProperty {
     /**
      * Number of Slot Delivery Worker threads that should be started.
      */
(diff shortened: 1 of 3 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
```diff
@@ -432,21 +432,21 @@
      * (- slots which have a aged more than 'messageAccumulationTimeout')
      */
     PERFORMANCE_TUNING_SUBMIT_SLOT_TIMER_PERIOD (
-            "performanceTuning/slots/timerPeriod", "3000", Integer.class, PERFORMANCE_TUNING_SUBMIT_SLOT_TIMEOUT),
+            "performanceTuning/slots/maxSubmitDelay", "3000", Integer.class, PERFORMANCE_TUNING_SUBMIT_SLOT_TIMEOUT),
```
sdkottegoda: Don't we need to change the name of the variabl to be the same as the configuration?

vinok88:
Don't we have any public Jira for this?

PR #822, opened 2017-02-22
Refactor changes in #820

sdkottegoda:
Refactored changes that were done in https://github.com/wso2/andes/pull/820/files

Message deletion from the database was moved inside the storage queue.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessageHandler.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessageHandler.java
@@ -319,11 +319,47 @@ public class MessageHandler {
     }
 
     /**
+     * Removes all the messages from read buffer, deletes the slots and deletes all messages from persistent storage.
+     *
+     * @return the number of messages that were deleted from the store
+     */
+    public int purgeMessagesOfQueue() throws AndesException {
+
+        try {
+            /*
+             * Clear all slots assigned to the Queue. This should ideally stop
+             * any messages being buffered during the purge. This call clears all slot associations
+             * for the queue in all nodes calling to coordinator node via Thrift protocol (could take time).
+             */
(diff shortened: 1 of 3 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/StorageQueue.java
```diff
@@ -414,17 +414,28 @@ public int clearMessagesReadToBufferForDelivery() {
...
-    public int purgeMessages() {
+    public int purgeMessages() throws AndesException {
         lastPurgedTimestamp = System.currentTimeMillis();
         log.info("Purging messages of queue " + name);
         return messageHandler.purgeMessagesOfQueue();
     }
 
     /**
+     * Purge the messages in the queue. This will remove all the messages from read buffer.
+     *
```
abeykoon: What about comment "Purge only in-memory messages of the queue. This will remove all the messages from read buffer" ?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/MessageHandler.java
```diff
@@ -319,11 +319,47 @@ public int clearReadButUndeliveredMessages() {
...
+        try {
+            /*
+             * Clear all slots assigned to the Queue. This should ideally stop
+             * any messages being buffered during the purge. This call clears all slot associations
+             * for the queue in all nodes calling to coordinator node via Thrift protocol (could take time).
+             */
+            MessagingEngine.getInstance().getSlotCoordinator().clearAllActiveSlotRelationsToQueue(queueName);
+
+            //clear all messages read to memory and return the slots
+            clearReadButUndeliveredMessages();
```
abeykoon: What about purgeInMemoryMessagesOfQueue() instead?

PR #828, opened 2017-02-28
MB-1892 issue fixed.

indikasampath2000:
Null subscription assigned if the subscriber connection is not a local. But when rendering nondurable subscriptions, we are fetching all subscriptions. Therefore, the condition added to check whether the subscription is local or not and based on that protocol queue name will retrieve. Also, NullSubscription class modified to return UnsupportedOperationException.

NullPointerException didn't occur following the given steps.

Public JIRA - https://wso2.org/jira/browse/MB-1892


modules/andes-core/broker/src/main/java/org/wso2/andes/server/information/management/SubscriptionManagementInformationMBean.java
```diff
@@ -55,6 +55,11 @@
     private static final String NOT_AVAILABLE = "N/A";
 
     /**
+     * Used to non local subscription protocol queue name
+     */
+    private static final String EMPTY_NAME = "";
```
sdkottegoda: Can't we use the same NOT_AVAILABLE string defined above for this. It will make the code look clean :)

indikasampath2000: +1

PR #829, opened 2017-02-28
Propagate DtxDbWriter related errors to the StateEventHandler

Asitha:
- Propagate dtx.commit/ditx.rollback related event handling
  errors to the StateEventHandler.


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/dtx/DtxBranch.java
```diff
@@ -423,12 +423,18 @@ public void writeToDbOnCommit() throws AndesException {
...
-     * {@inheritDoc}
+     * Update the state of the transaction and respond to the client on dtx.commt and dtx.rollback events
+     *
+     * @param event {@link InboundEventContainer}
+     * @throws AndesException
      */
-    @Override
-    public void updateState() throws AndesException {
+    public void updateState(InboundEventContainer event) throws AndesException {
         if (updateSlotCommand != null) {
```
a5anka: Can this be null?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/disruptor/inbound/InboundEventContainer.java
```diff
@@ -120,6 +121,30 @@ public DtxBranch getDtxBranch() {
     }
 
     /**
+     * Set the error occurred while processing the event using Disruptor event handler
+     * @param error Throwable
+     */
+    public void setError(Throwable error) {
+        this.error = error;
```
a5anka: If the error is not null, better to print an error log.

Asitha: This is called through `LogExceptionHandler` which logs each exception.

PR #841, opened 2017-03-07
MB-1901 issue fixed.

indikasampath2000:
If durable queue/topic subscriber got an error while sending the message, then it is routed to the DLC.
But in the current implementation, if any error occurred, the message got deleted instead of routing to DLC.
The reason is we are setting internalErrorOccurred to true. This PR modified routeMessageToDLC method
by removing the above-mentioned parameter. Now messages will be routed to DCL in a case of error
situation while sending. Only nondurable topic subscriber messages will be discarded.


indikasampath2000:
Public JIRA - https://wso2.org/jira/browse/MB-1901

PR #843, opened 2017-03-08
[MB-1909] fixed issue in safe zone update

sdkottegoda:
Fixes the issue of the safe zone not getting updated in a fail-over scenario reported at https://wso2.org/jira/browse/MB-1909

The issue was due to the fact that 'slotSubmitLoopSkipCount' in SlotMessageCounter(which is compared with SLOT_SUBMIT_LOOP_SKIP_COUNT_THRESHOLD to update the safezone) is not being reset when a connection exception occurs when the coordinator goes down. 

Removed 'slotSubmitLoopSkipCount' itself since the operation to update the safe zone already happens as a scheduled task.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotMessageCounter.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotMessageCounter.java
@@ -68,14 +68,7 @@ public class SlotMessageCounter implements StoreHealthListener {
     private final int slotWindowSize;
     private long currentSlotDeleteSafeZone;
 
-    /**
-     * Keep track of how many update loops are skipped without messages.
-     */
-    private int slotSubmitLoopSkipCount;
-
     private SlotCoordinator slotCoordinator;
-
-    private static final int SLOT_SUBMIT_LOOP_SKIP_COUNT_THRESHOLD = 10;
 
     /**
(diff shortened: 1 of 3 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/slot/SlotMessageCounter.java
```diff
@@ -361,39 +353,35 @@ private void updateCoordinatorWithTimedOutSlots(Set<Map.Entry<String, Long>> slo
...
-                } catch (ConnectionException e) {
-                    log.error("Error while sending slot deletion safe zone update", e);
+                if (log.isDebugEnabled()) {
+                    log.debug("Updating coordinator with local safe zone " + evaluatedSafeZone);
                 }
+
+                submitCurrentSafeZone(evaluatedSafeZone);
+                currentSlotDeleteSafeZone = evaluatedSafeZone;
+            } catch (ConnectionException e) {
+                log.error("Error while sending slot deletion safe zone update", e);
```
lankavitharana: isn't it better to add e.getmessage() as part of the log itself?

sdkottegoda: Since the error itself will anyway get printed, adding that will duplicate the message.

PR #844, opened 2017-03-08
XA transactions fail-over testing fixes

a5anka:
Suggestions for a different failover policy name is welcomed. :smiley: Currently we use the name "onetime". Broker URL will look something like below.

```
connectionfactory.QueueConnectionFactory = amqp://admin:admin@clientID/carbon?failover='onetime'&cyclecount='2'&brokerlist='tcp://localhost:5673?retries='5'&connectdelay='50';tcp://localhost:5674?retries='5'&connectdelay='50';tcp://localhost:5675?retries='5'&connectdelay='50''
```

### DIFF
--- a/modules/andes-core/client/src/main/java/org/wso2/andes/client/XASession_9_1.java
+++ b/modules/andes-core/client/src/main/java/org/wso2/andes/client/XASession_9_1.java
@@ -231,6 +231,7 @@ class XASession_9_1 extends AMQSession_0_8 implements XASession, XAQueueSession,
      * @throws AMQException      if server sends back a error response
      */
     public XaStatus forget(Xid xid) throws FailoverException, AMQException, XAException {
+
         throwErrorIfClosed();
 
         DtxForgetBody dtxForgetBody = methodRegistry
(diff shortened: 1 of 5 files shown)


modules/andes-core/client/src/main/java/org/wso2/andes/client/XASession_9_1.java
```diff
@@ -154,7 +157,10 @@ XaStatus startDtx(Xid xid, int flag) throws FailoverException, AMQException {
      * @throws FailoverException when a connection issue is detected
      * @throws AMQException      when an error is detected in AMQ state manager
      */
-    public XaStatus endDtx(Xid xid, int flag) throws FailoverException, AMQException {
+    public XaStatus endDtx(Xid xid, int flag) throws FailoverException, AMQException, XAException {
+
+        throwErrorIfClosed();
+
         if (LOGGER.isDebugEnabled()) {
```
sdkottegoda: Wouldn't it be better if we move the up?

a5anka: I don't think so since we are not sending the dtx.end if the session is closed. If we move it up we can have a log entry which says "Sending dtx.end for channel ..." but actually we never tried doing that.

PR #861, opened 2017-03-24
[MB-1919] fixed issue of messages being buffered twice when failing over

sdkottegoda:
Fixes the issues reported at:
1. https://wso2.org/jira/browse/MB-1919
2. https://wso2.org/jira/browse/MB-1920

Includes the following changes:
1. Added channel flow suspend/resume to broker
This is sent from andes slient prior to and after recovery respectively. Message will not be buffered to the susbscription that is associated with the channel id throughout the time that the channel is suspended. 

2. A message which has to delivery statuses as RECOVERED and ACKED from 2 different channel will get  deleted upon the ack.


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/DeliverableAndesMetadata.java
```diff
@@ -482,6 +482,10 @@ private boolean isMarkAsAcked() {
             if (null != messageStatus && messageStatus.equals(ChannelMessageStatus.CLIENT_REJECTED)) {
                 continue;
             }
+            //the message could be recovered by one client and acknowledged by another, hence needs to be ignored
+            if (null != messageStatus && messageStatus.equals(ChannelMessageStatus.RECOVERED)) {
```
a5anka: Do we need to check null != messageStatus at each if? Can't we only check it at the top?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/StorageQueue.java
```diff
@@ -141,6 +146,7 @@ public StorageQueue(String queueAsStr) {
             }
         }
         this.boundedSubscriptions = new ArrayList<>(1);
+        this.boundUnSuspendedSubscriptions = new ArrayList<>(1);
```
a5anka: Same code use in line 114. Can't we use a common place for initializing?

modules/andes-core/client/src/main/java/org/wso2/andes/client/AMQSession.java
```diff
@@ -1855,7 +1853,7 @@ public void recover() throws JMSException
             // Set inRecovery to false before you start message flow again again.            
             _inRecovery = false; 
             
-            if (!isSuspended)
+            if (_suspended)
```
a5anka: In the previous code we are checking the negation of suspend. Why have we changed that?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/disruptor/inbound/InboundChannelFlowEvent.java
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     * Supported state events.
+     */
+    public enum EventType {
+
+        /**
+         * channel flow event signals the channel to be suspended/resumed.
+         */
+        CHANNEL_FLOW_EVENT
```
a5anka: Do we need an event type here? InboundChannelFlowEvent only has a single event type

PR #862, opened 2017-03-27
[MB-1895] - Fix Message ordering issue with local JMS transactions

hastef88:
Client side - 

When rolling back messages not yet seen by the application, iteration of the DelayedQueue will cause jumbling of message order. Therefore used a while -> take/pull loop. 

After all messages are rejected, client side must wait until all rejected messages arrive in order to release the rollback lock. Used a CountdownLatch for this purpose. 

Server Side : 

Server must know the last seen message ID by the application, and should only increase the message.deliveryCount if the requeing message ID <= to the lastSeenMessageID. Wrote a custom RollbackWithContext frame for this (to replace AMQP "rollback" frame). So, when using JMS transactions with our java andes client, this new frame will be used. When using any other AMQP client, the existing "rollback" flow will execute. 

* In amqp-0-91 section 2.2.9, The rollback behavior does not guarantee redelivery of the messages, but relies on client's acknowledgement. In JMS, the message order should be guaranteed. Hence, this new frame. 

While the server is handling the rollback event from client side, no new messages should be dispatched to the client. In our case, the MessageFlusher will asynchronously queue such new messages regardless of rollback. Used a kind of a delivery rule to avoid new messages being queued until rollback is complete.

### DIFF
--- a/modules/andes-core/client/src/main/java/org/wso2/andes/client/BasicMessageConsumer.java
+++ b/modules/andes-core/client/src/main/java/org/wso2/andes/client/BasicMessageConsumer.java
@@ -42,6 +42,7 @@ import java.util.List;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.DelayQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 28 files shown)


modules/andes-core/client/src/main/java/org/wso2/andes/client/BasicMessageConsumer.java
```diff
@@ -1217,4 +1257,56 @@ public void run() {
...
+    public void waitUntilAllRejectsReceived() {
+        if (lastRejectedDeliveryTag > -1) {
+            if (_logger.isDebugEnabled()) {
+                _logger.debug("Starting to wait for rejects to come back ! " +
+                        "lastRejectedDeliveryTag : " +
+                        lastRejectedDeliveryTag + " lastRejectedJMSMessageID : " + lastRejectedJMSMessageID);
+            }
+
+            try {
+                allRejectsReceivedCountdown.await(20, TimeUnit.SECONDS);
```
hastef88: To configure or not to configure..

PR #863, opened 2017-03-27
MB-1895 - Fix #1 - message order during transaction rollback. 

hastef88:
When rolling back messages not yet seen by the application, iteration of the DelayedQueue will cause jumbling of message order. Therefore used a while -> take/pull loop.

*Further ordering issues still occur if the subscriber is consuming messages without any delay. These will be fixed later.

### DIFF
--- a/modules/andes-core/client/src/main/java/org/wso2/andes/client/BasicMessageConsumer.java
+++ b/modules/andes-core/client/src/main/java/org/wso2/andes/client/BasicMessageConsumer.java
@@ -1062,6 +1062,7 @@ public abstract class BasicMessageConsumer<U> extends Closeable implements Messa
     public void rollbackPendingMessages()
     {
 
+
         if (_synchronousQueue.size() > 0)
         {
             _logger.debug("dest="+ _destination.getQueueName()+  " rolling back messages [" + _synchronousQueue.size() + "]");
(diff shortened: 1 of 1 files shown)


modules/andes-core/client/src/main/java/org/wso2/andes/client/BasicMessageConsumer.java
```diff
@@ -1072,51 +1073,45 @@ public void rollbackPendingMessages()
...
-            while (iterator.hasNext())
-            {
 
-                Object o = iterator.next().getObject();
-
-                if (o instanceof AbstractJMSMessage)
-                {
-                    try {
+            while (_synchronousQueue.size() > 0) {
+                Object o = _synchronousQueue.peek().getObject();
```
Asitha: Should this be a variable like `message`

modules/andes-core/client/src/main/java/org/wso2/andes/client/BasicMessageConsumer.java
```diff
@@ -1072,51 +1073,45 @@ public void rollbackPendingMessages()
...
-                                }
+            while (_synchronousQueue.size() > 0) {
+                Object message = _synchronousQueue.peek().getObject();
+                try {
+                    if (message instanceof AbstractJMSMessage) {
+                        if ((lastRollbackedMessageTimestamp > 0) && ((AbstractJMSMessage) message).getJMSRedelivered() &&
+                                (lastRollbackedMessageTimestamp > ((AbstractJMSMessage) message).getJMSTimestamp())) {
+                            if (_logger.isDebugEnabled()) {
+                                _logger.debug("Did not remove message " + printMessage((AbstractJMSMessage) message) +
+                                        " since its new relative to the rollback point." +
```
sdkottegoda: '+' operator should be at the beginning of a line

PR #867, opened 2017-03-28
Ei compatibility fixes

hastef88:
Porting fix : https://github.com/wso2/andes/pull/777

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/AndesConfigurationManager.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/AndesConfigurationManager.java
@@ -85,7 +85,7 @@ public class AndesConfigurationManager {
     /**
      * Namespace and attribute used by secure vault to identify encrypted properties.
      */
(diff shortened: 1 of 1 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/AndesConfigurationManager.java
```diff
@@ -110,8 +111,9 @@
      * Main file path of configuration files into which all other andes-specific config files (if any)
      * should be linked.
      */
-    private static final String ROOT_CONFIG_FILE_PATH = System.getProperty(ServerConstants.CARBON_HOME) +
-            "/repository/conf/";
+    public static final String CARBON_CONFIG_DIR_PATH = "carbon.config.dir.path";
+    public static final String CARBON_HOME = "carbon.home";
+    private static String componentsPath = System.getProperty(CARBON_CONFIG_DIR_PATH);
```
Asitha: Would it be better, if we change the name to  `configPath`?

PR #877, opened 2017-03-31
[MB-1922] Fix : NPE when restoring empty messages from DLC queue

manoramahp:
This PR contains the fix for the NPE thrown when restoring messages having empty content from the DLC queue.
Fix : https://wso2.org/jira/browse/MB-1922

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/server/information/management/QueueManagementInformationMBean.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/server/information/management/QueueManagementInformationMBean.java
@@ -1136,9 +1136,11 @@ public class QueueManagementInformationMBean extends AMQManagedObject implements
 
             messagesToRemove.add(metadata);
             // Update Andes message with all the chunk details
-            List<AndesMessagePart> messageParts = messageContent.get(messageId);
-            for (AndesMessagePart messagePart : messageParts) {
-                andesMessage.addMessagePart(messagePart);
+            if(!messageContent.isEmpty()) {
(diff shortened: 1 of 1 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/server/information/management/QueueManagementInformationMBean.java
```diff
@@ -1136,9 +1136,11 @@ private int moveMessagesFromDLCToNewDestination(List<Long> messageIds, String so
 
             messagesToRemove.add(metadata);
             // Update Andes message with all the chunk details
-            List<AndesMessagePart> messageParts = messageContent.get(messageId);
-            for (AndesMessagePart messagePart : messageParts) {
-                andesMessage.addMessagePart(messagePart);
+            if(!messageContent.isEmpty()) {
```
abeykoon: Can message content be empty? In the test case an empty message was sent?

manoramahp: When testing an empty string was sent as the message and observed the messageContent as an empty map (ie. {}).

sdkottegoda: Theoritically, if the messageContent map is empty it will anyway not go inside for loop, right?

manoramahp: NPE is thrown at : `List<AndesMessagePart> messageParts = messageContent.get(messageId);` before go inside loop.

PR #879, opened 2017-04-02
Fix cluster notification missing in the RDMBS coordination

indikasampath2000:
RDBMS coordination read notifications from MB_CLUSTER_EVENT and MB_MEMBERSHIP tables.
After selecting notifications, those are deleted by the destined node. This would
lead delete unread notifications in a high concurrent scenario. The fix is select
notifications with event id and then delete notifications by event id.
Additionally, DEBUG logs modified with more details in RDBMSBasedNotificationAgentImpl
and HazelcastBasedNotificationAgentImpl.

Public JIRAs - https://wso2.org/jira/browse/MB-1839, https://wso2.org/jira/browse/MB-1881

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSAndesContextStoreImpl.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSAndesContextStoreImpl.java
@@ -2555,16 +2555,19 @@ public class RDBMSAndesContextStoreImpl implements AndesContextStore {
             preparedStatement.setString(1, nodeID);
             resultSet = preparedStatement.executeQuery();
 
+            clearMembershipEvents = connection.prepareStatement(RDBMSConstants.PS_CLEAN_MEMBERSHIP_EVENTS_FOR_EVENT_ID);
+
             while (resultSet.next()) {
                 MembershipEvent membershipEvent = new MembershipEvent(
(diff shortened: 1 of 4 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSAndesContextStoreImpl.java
```diff
@@ -2647,7 +2650,7 @@ public void storeClusterNotification(List<String> clusterNodes, String originate
 
         Connection connection = null;
         PreparedStatement preparedStatement = null;
```
sdkottegoda: Shall we rename this as well?

modules/andes-core/broker/src/main/java/org/wso2/andes/store/rdbms/RDBMSConstants.java
```diff
@@ -1049,6 +1053,10 @@
             "DELETE FROM " + MEMBERSHIP_TABLE
             + " WHERE " + NODE_ID + "=?";
 
+    protected static final String PS_CLEAN_MEMBERSHIP_EVENTS_FOR_EVENT_ID =
+            "DELETE FROM " + MEMBERSHIP_TABLE
```
sdkottegoda: Shall we follow the coding style to align with other statements

PR #880, opened 2017-04-03
[MB-1921] Add a scheduled close task to XAConnectionImpl

a5anka:
This to avoid stale connections when a commit or rollback get delayed for a long time.

https://wso2.org/jira/projects/MB/issues/MB-1921

### DIFF
--- a/modules/andes-core/client/src/main/java/org/wso2/andes/client/XAConnectionImpl.java
+++ b/modules/andes-core/client/src/main/java/org/wso2/andes/client/XAConnectionImpl.java
@@ -23,6 +23,9 @@ import org.wso2.andes.AMQException;
 import org.wso2.andes.jms.ConnectionURL;
 
 import java.util.ArrayList;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 import javax.jms.IllegalStateException;
 import javax.jms.JMSException;
 import javax.jms.XAConnection;
@@ -43,21 +46,39 @@ public class XAConnectionImpl extends AMQConnection implements XAConnection, XAQ
     private static final Logger LOGGER = LoggerFactory.getLogger(XAConnectionImpl.class);
 
     /**
+     * After this delay(in seconds), the connection will be closed even if a commit or rollback is not received. This
+     * is done to avoid stale connections to broker.
+     */
+    private static final int CONNECTION_CLOSE_TIMEOUT = 60;
+
+    /**
      * Keep track of active XA sessions
      */
     private final ArrayList<XASession_9_1> xaSessions = new ArrayList<>();
+
+    /**
+     * Executor service used to schedule connection close task.
+     */
+    private final ScheduledExecutorService scheduledExecutor;
 
     /**
      * Indicate if the connection.close is called before
      */
     private boolean connectionCloseSignaled = false;
 
+    /**
+     * Hold the future object of the scheduled close task
+     */
+    private ScheduledFuture<?> connectionCloseFuture;
+
     //-- constructor
     /**
      * Create a XAConnection from a connectionURL
      */
-    XAConnectionImpl(ConnectionURL connectionURL, SSLConfiguration sslConfig) throws AMQException {
+    XAConnectionImpl(ConnectionURL connectionURL, SSLConfiguration sslConfig,
+            ScheduledExecutorService scheduledExecutor) throws AMQException {
         super(connectionURL, sslConfig);
+        this.scheduledExecutor = scheduledExecutor;
     }
 
     //-- interface XAConnection
@@ -65,7 +86,7 @@ public class XAConnectionImpl extends AMQConnection implements XAConnection, XAQ
      * Creates an XASession.
      *
      * @return A newly created XASession.
-     * @throws JMSException If the XAConnectiono fails to create an XASession due to
+     * @throws JMSException If the XAConnection fails to create an XASession due to
      *                      some internal error.
      */
     public synchronized XASession createXASession() throws JMSException {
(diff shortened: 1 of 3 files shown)


modules/andes-core/client/src/main/java/org/wso2/andes/client/XAConnectionImpl.java
```diff
@@ -43,29 +46,47 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(XAConnectionImpl.class);
 
     /**
+     * After this delay(in seconds), the connection will be closed even if a commit or rollback is not received. This
+     * is done to avoid stale connections to broker.
+     */
+    private static final int CONNECTION_CLOSE_TIMEOUT = 60;
```
hastef88: should this be configurable ?

a5anka: We also need to decide on the default value. Is 60 seconds enough?

modules/andes-core/client/src/main/java/org/wso2/andes/client/XAConnectionImpl.java
```diff
@@ -43,29 +46,47 @@
...
+
+    /**
      * Indicate if the connection.close is called before
      */
     private boolean connectionCloseSignaled = false;
 
+    /**
+     * Hold the future object of the scheduled close task
+     */
+    private ScheduledFuture<?> scheduledFuture;
```
hastef88: use a more meaningful name maybe ? staleConnectionCloseTask ? connectionCloseScheduler ?

a5anka: How about `connectionCloseFuture`?

hastef88: +1

modules/andes-core/client/src/main/java/org/wso2/andes/client/XAConnectionImpl.java
sdkottegoda: Can we test the existing setup and see the average time it takes for a commit to get received? We should be able to get an idea from that.

modules/andes-core/client/src/main/java/org/wso2/andes/client/XAConnectionImpl.java
```diff
@@ -145,9 +178,35 @@ synchronized void deregisterSession(XASession_9_1 xaSession) {
      *
      * @throws JMSException if there was an error closing the physical connection
      */
-    synchronized void internalClose() throws JMSException {
+    void internalClose() throws JMSException {
+        boolean closeSuccessful = closeIfNoActiveSessions();
+        if (closeSuccessful) {
```
hemikak: Do we need to assign it to closeSuccessful ? Cant we use the closeIfNoActiveSessions() itself as the condition ?

modules/andes-core/client/src/main/java/org/wso2/andes/client/XAConnectionImpl.java
```diff
@@ -124,6 +145,18 @@ public synchronized void close() throws JMSException {
...
+                    public void run() {
+                        try {
+                            LOGGER.error("Closing XAConnection after waiting " + CONNECTION_CLOSE_TIMEOUT
+                                                 + " seconds for a commit or rollback");
+                            closeIfNoActiveSessions();
+                        } catch (JMSException e) {
+                            LOGGER.error("Error occurred while closing the XAConnection after close timeout");
+                        }
+                    }
+                }, CONNECTION_CLOSE_TIMEOUT, TimeUnit.SECONDS);
```
hemikak: Cant we stick to milliseconds ? Most of our configurations are so, isnt it ?

modules/andes-core/client/src/main/java/org/wso2/andes/client/XAConnectionImpl.java
```diff
@@ -43,29 +46,47 @@
...
+
     //-- constructor
     /**
      * Create a XAConnection from a connectionURL
      */
-    XAConnectionImpl(ConnectionURL connectionURL, SSLConfiguration sslConfig) throws AMQException {
+    XAConnectionImpl(ConnectionURL connectionURL, SSLConfiguration sslConfig,
+            ScheduledExecutorService scheduledExecutor) throws AMQException {
         super(connectionURL, sslConfig);
+        this.scheduledExecutor = scheduledExecutor;
```
sdkottegoda: All the usages of scheduled executor are within XAConnectionImpl. So why do we need to pass it as a parameter to the constructor?

modules/andes-core/client/src/main/java/org/wso2/andes/client/XAConnectionImpl.java
a5anka: Currently, it happens instantaneously

modules/andes-core/client/src/main/java/org/wso2/andes/client/XAConnectionImpl.java
a5anka: Scheduled Executor is shared to make the thread count minimal.

modules/andes-core/client/src/main/java/org/wso2/andes/client/XAConnectionImpl.java
a5anka: A local variable was introduced to make it more readable. The compiler might do the exaplained optimization.

modules/andes-core/client/src/main/java/org/wso2/andes/client/XAConnectionImpl.java
a5anka: I count this as 1 vote for a config. :smile:  I do not think we need to configure this timeout in Millisecond's level since this is not a performance tuning parameter rather a fail safe parameter.

modules/andes-core/client/src/main/java/org/wso2/andes/client/XAConnectionImpl.java
hemikak: 👍

PR #881, opened 2017-04-04
[MB-1920] fixed issue of messages not being deleted in fail-over

sdkottegoda:
Addresses the issue reported at https://wso2.org/jira/browse/MB-1920

A message which has 2 delivery statuses as RECOVERED and ACKED from 2 different channel will get deleted upon the ack.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/DeliverableAndesMetadata.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/DeliverableAndesMetadata.java
@@ -474,17 +474,26 @@ public class DeliverableAndesMetadata extends AndesMessageMetadata {
         for (Map.Entry<UUID, ChannelInformation> channelInfoEntry : channelDeliveryInfo.entrySet()) {
             ChannelMessageStatus messageStatus = channelInfoEntry.getValue().getLatestMessageStatus();
 
-            //if channel is closed ignore it from considering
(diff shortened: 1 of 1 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/DeliverableAndesMetadata.java
```diff
@@ -474,17 +474,26 @@ private boolean isMarkAsAcked() {
...
-            //if channel is closed ignore it from considering
-            if (null != messageStatus && messageStatus.equals(ChannelMessageStatus.CLOSED)) {
-                continue;
-            }
-            //if message is rejected by client repeatedly ignore it from considering
-            if (null != messageStatus && messageStatus.equals(ChannelMessageStatus.CLIENT_REJECTED)) {
-                continue;
-            }
-            if (null == messageStatus || !messageStatus.equals(ChannelMessageStatus.ACKED)) {
+            if (null == messageStatus) {
```
a5anka: @abeykoon are there any scenarios where the messageStatus can be null here?

abeykoon: There is. 

ChannelInformation object has the list of channel message status. ChannelInformation is created when a message is "scheduled" for a subscriber. First status added is "DISPATCHED" when disruptor write message to wire. Thus when a message is in the disruptor channel message status list is empty. Thus null check is needed.

PR #887, opened 2017-04-06
Add message tracing for distributed transactions 

a5anka:


modules/andes-core/broker/src/main/java/org/wso2/andes/tools/utils/MessageTracer.java
```diff
@@ -396,72 +399,53 @@ public static void trace(DeliverableAndesMetadata metadata, Slot slot, String de
...
          * Set message id to the trace
+         *
          * @param id Message id
          * @return TraceBuilder
          */
         TraceBuilder setMessageId(long id) {
-            checkAndAddTraceHeader();
-            messageContent.append(MESSAGE_ID);
-            messageContent.append(id);
+            messageContent.append(FIELD_SEPARATOR).append(MESSAGE_ID).append(id);
```
Asitha: Would this have a trace message like `Message {  , id: 43543534435 }`  if we add the separator without checking?

a5anka: Yes, Will fix that

PR #888, opened 2017-04-06
Fix MB-1925

a5anka:
The issue is fixed by using AMQChannel's ID for transaction association.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/dtx/DtxBranch.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/dtx/DtxBranch.java
@@ -16,7 +16,6 @@
 package org.wso2.andes.kernel.dtx;
 
 import org.apache.log4j.Logger;
-import org.wso2.andes.api.Message;
 import org.wso2.andes.kernel.AndesAckData;
 import org.wso2.andes.kernel.AndesChannel;
 import org.wso2.andes.kernel.AndesException;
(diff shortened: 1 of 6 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/dtx/DtxBranch.java
```diff
@@ -219,7 +218,7 @@ boolean disassociateSession(long sessionID) {
      * @param sessionID session identifier of the session
      * @return True if there is a matching suspended entry
      */
-    boolean resumeSession(long sessionID) {
+    boolean resumeSession(UUID sessionID) {
```
pamod: How about sticking with the convention of "sessionId" in camelcase ?

Asitha:
Shall we add a test case for this?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/dtx/DtxBranch.java
```diff
@@ -54,7 +54,7 @@
     /**
      * Session Id of a session that is recovered from storage
      */
-    public static final int RECOVERY_SESSION_ID = -1;
+    public static final UUID RECOVERY_SESSION_ID = UUID.randomUUID();
```
Asitha: Can you explain how a random value replaces the previous constant value?

a5anka: Actually, we don't need this constant. It is used as the session id for recovered branches.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/dtx/DtxBranch.java
a5anka: :+1:

PR #889, opened 2017-04-07
Fix Message Tracer field separator issue

a5anka:


modules/andes-core/broker/src/main/java/org/wso2/andes/tools/utils/MessageTracer.java
```diff
@@ -413,18 +415,28 @@ public static void trace(DeliverableAndesMetadata metadata, Slot slot, String de
...
-         * String builder for trace message
+         * Used at the end of the trace message
          */
-        private StringBuilder messageContent;
+        static final String POSTFIX = " } ";
+
+        /**
+         * Hold the list of field items
+         */
+        private final ArrayList<String> fields;
```
pamod: Can we add the generic type List<String> instead ?

PR #895, opened 2017-04-18
[MB-1927] Fix failing recovery tests

sdkottegoda:
Fixes issues in message recovery which caused the failure of test cases. Tracked at https://wso2.org/jira/browse/MB-1927

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscription.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscription.java
@@ -105,7 +105,7 @@ public class AndesSubscription {
      *
      * @return a clone of the subscription that is not suspended or obsolete
      */
-    public AndesSubscription cloneAndResetSubscription() {
+    private AndesSubscription cloneAndResetSubscription() {
         return new AndesSubscription(subscriptionId, storageQueue, protocolType,
                 subscriberConnection.createFreshClone());
     }
@@ -246,7 +246,7 @@ public class AndesSubscription {
      * @throws AndesException on a message re-schedule issue
      * @return DeliverableAndesMetadata reference of rejected message
      */
-    public synchronized DeliverableAndesMetadata onMessageReject(long messageID, boolean reQueue)
+    public DeliverableAndesMetadata onMessageReject(long messageID, boolean reQueue)
             throws AndesException {
         DeliverableAndesMetadata rejectedMessage = subscriberConnection.onMessageReject(messageID);
         //Adding metrics meter for ack rate
@@ -281,7 +281,7 @@ public class AndesSubscription {
     /**
      * Recover messages of the subscriber. Buffers all sent but un-acknowledged messages back to the queue it is bound.
      */
-    public synchronized void recoverMessages() throws AndesException {
+    public void recoverMessages() throws AndesException {
 
         //stop message delivery to the channel
         prepareRecover();
@@ -314,15 +314,6 @@ public class AndesSubscription {
                 }
             }
         }
-    }
-
-    /**
-     * Removes a message that is buffered to be scheduled given the message id.
-     *
-     * @param messageId id of the message to be removed from the buffer
-     */
-    public void removeMessageFromBuffer(long messageId) {
-        storageQueue.removeMessageFromBuffer(messageId);
     }
 
     /**
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscriptionManager.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscriptionManager.java
@@ -632,7 +632,9 @@ public class AndesSubscriptionManager implements NetworkPartitionListener, Store
         if (subIterator.hasNext()) {
             return subIterator.next();
         } else {
-            log.warn("No subscription found for channel ID " + channelID);
+            if (log.isDebugEnabled()) {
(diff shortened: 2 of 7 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/server/handler/ChannelFlowHandler.java
```diff
@@ -81,6 +81,6 @@ public void onException(Exception exception) {
 
             }
         });
-        logger.info(" Channel flow: " + body.getActive() + " set for channel: " + channel.getId());
+        logger.debug(" Channel flow: " + body.getActive() + " set for channel: " + channel.getId());
     }
```
manoramahp: Shall we call logger.isDebugEnabled() before logger.debug() since the log message contains some string concatenation?

ramith: +1

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscription.java
```diff
@@ -246,7 +246,7 @@ public void onMessageAck(long messageID) throws AndesException {
      * @throws AndesException on a message re-schedule issue
      * @return DeliverableAndesMetadata reference of rejected message
      */
-    public synchronized DeliverableAndesMetadata onMessageReject(long messageID, boolean reQueue)
```
a5anka: Don't we need to synchronize this method?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscription.java
```diff
@@ -281,7 +281,7 @@ public synchronized DeliverableAndesMetadata onMessageReject(long messageID, boo
     /**
      * Recover messages of the subscriber. Buffers all sent but un-acknowledged messages back to the queue it is bound.
      */
-    public synchronized void recoverMessages() throws AndesException {
```
a5anka: Don't we need to synchronize this method?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscriptionManager.java
```diff
@@ -632,7 +632,9 @@ public AndesSubscription getSubscriptionByProtocolChannel(UUID channelID) {
         if (subIterator.hasNext()) {
             return subIterator.next();
         } else {
-            log.warn("No subscription found for channel ID " + channelID);
+            if (log.isDebugEnabled()) {
+                log.debug("No subscription found for channel ID " + channelID);
```
a5anka: Can this happen frequently? If not we should keep it as a warn.

modules/andes-core/broker/src/main/java/org/wso2/andes/server/AMQChannel.java
```diff
@@ -299,6 +299,10 @@ public ConfigStore getConfigStore()
      */
     public void resendRecoveredMessages(DisruptorEventCallback recoverOKCallback) {
         try {
+
+            //messages should be removed from the unacknowledged map upfront
+            _unacknowledgedMessageMap.clear();
```
a5anka: I thinks it is better to call `_unacknowledgedMessageMap.cancelAllMessages()` which also reset the `_unackedSize`

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscription.java
sdkottegoda: It was required ealrier because  different outbound message trackers were being used in each invocation earlier. Now that the same is used, synchronication is no longer required.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscription.java
sdkottegoda: Same as above

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscriptionManager.java
sdkottegoda: Yes, each time a subscriber is connected, this happens. Besides it's not the correct place to print the log. This is just a method which returns suscriptions by the channel id. The one who actually knows if it's a valid channel id is the upper layer calling this. WDYT?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscription.java
a5anka: :+1:

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscriptionManager.java
a5anka: Is it correct to call the method`getSubscriptionByProtocolChannel()` for a non existing subscription?

sdkottegoda: We get a channel flow request to a channel for which a subscription is not yet registered. How can we handle that?

PR #897, opened 2017-04-19
Fix MB-1929

malakaganga:
Rejection was send upon closing the consumer hence separated the recovery in Autoack mode.

### DIFF
--- a/modules/andes-core/client/src/main/java/org/wso2/andes/client/BasicMessageConsumer.java
+++ b/modules/andes-core/client/src/main/java/org/wso2/andes/client/BasicMessageConsumer.java
@@ -630,11 +630,7 @@ public abstract class BasicMessageConsumer<U> extends Closeable implements Messa
                         // If the session is open or we are in the process
                         // of closing the session then send a cance
                         // no point otherwise as the connection will be gone
-                        if (!_session.isClosed() || _session.isClosing())
-                        {
-                            if(null != _session._dispatcher) {
-                                _session._dispatcher.rejectPending(this);
-                            }
+                        if (!_session.isClosed() || _session.isClosing()) {
                             sendCancel();
                             cleanupQueue();
                         }


modules/andes-core/client/src/test/java/org/wso2/andes/unit/message/TestAMQSession.java
```diff
@@ -47,6 +48,11 @@ public void acknowledgeMessage(long deliveryTag, boolean multiple)
 
     }
 
+    public void rejectRecoveryMessage(long deliveryTag, boolean requeue)
```
akalankapagoda: Can we explain what this is why this is not implemented.

modules/andes-core/client/src/test/java/org/wso2/andes/unit/message/TestAMQSession.java
```diff
@@ -47,6 +48,11 @@ public void acknowledgeMessage(long deliveryTag, boolean multiple)
...
+    {
+        throw new NotImplementedException();
+    }
+
```
a5anka: Code style is not correct. Shall we fix?

modules/andes-core/client/src/main/java/org/wso2/andes/client/BasicMessageConsumer.java
```diff
@@ -630,11 +630,7 @@ public void close(boolean sendClose) throws JMSException
                         // If the session is open or we are in the process
                         // of closing the session then send a cance
                         // no point otherwise as the connection will be gone
-                        if (!_session.isClosed() || _session.isClosing())
-                        {
-                            if(null != _session._dispatcher) {
-                                _session._dispatcher.rejectPending(this);
-                            }
+                        if (!_session.isClosed() || _session.isClosing()) {
```
pamod: Any reason for using '_' value with variables ?

malakaganga: This is variable is coming from Qpid code which is  a external dependency.

PR #898, opened 2017-05-01
Avoid generating a client ID if not provided

a5anka:


modules/andes-core/client/src/main/java/org/wso2/andes/client/url/URLParser.java
```diff
@@ -66,15 +65,7 @@ private void parseURL(String fullURL) throws URLSyntaxException
                 }
                 else
```
Asitha: Shall we get rid of this else block? We can add a comment saying, we are not auto generating a client id

PR #901, opened 2017-05-22
Add qpid independant Andes metadata implementation

Asitha:
- Implement a new Andes message metadata implementation
  independant of the existing qpid code.


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/AndesMessageMetadata.java
```diff
@@ -11,195 +11,322 @@
...
      *
-     * @see org.wso2.andes.kernel.AndesMessageMetadata#retain
-     * @param retain boolean retain flag
+     * @param destination  destination of the message
+     * @param protocolType {@link ProtocolType} of the message
      */
-    public void setRetain(boolean retain) {
-        this.retain = retain;
+    public AndesMessageMetadata(String destination, ProtocolType protocolType) {
+        this(0L, destination, protocolType);
```
hastef88: messageId = 0 here ?

PR #905, opened 2017-06-01
 Improve selectors using multiple cursors

riyafa:
Changes have been made to let AndesSubscription have a cursor and this cursor is used when filtering selectors for the messages. Also a sliding buffer has been used.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/FlowControlledQueueMessageDeliveryImpl.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/FlowControlledQueueMessageDeliveryImpl.java
@@ -119,7 +119,6 @@ public class FlowControlledQueueMessageDeliveryImpl implements MessageDeliverySt
 
                         //for queue messages and durable topic messages (as they are now queue messages)
                         // we only send to one selected subscriber if it is a queue message
-                        localSubscription.setCursor(currentCursor);
                         minCursor = Long.min(currentCursor, minCursor);
                         if (numOfCurrentMsgDeliverySchedules == 1) {
                             if (log.isDebugEnabled()) {
@@ -132,6 +131,7 @@ public class FlowControlledQueueMessageDeliveryImpl implements MessageDeliverySt
                         }
                         break;
                     }
+                    localSubscription.setCursor(currentCursor);
                 }
             }
             if (isRemoved) {
@@ -152,11 +152,12 @@ public class FlowControlledQueueMessageDeliveryImpl implements MessageDeliverySt
         }
         return sentMessageCount;
     }
+
     /**
      * Check if the message is unacked in at least one of the subscriptions.
      *
      * @param subscriptions The subscriptions to check in.
-     * @param messageId The message id of the message to see if it is unacked.
+     * @param messageId     The message id of the message to see if it is unacked.
      */
     private boolean checkIfUnacked(List<AndesSubscription> subscriptions, long messageId) {
         for (AndesSubscription subscription : subscriptions) {
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/FlowControlledQueueMessageDeliveryImpl.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/FlowControlledQueueMessageDeliveryImpl.java
@@ -18,19 +18,21 @@
 
 package org.wso2.andes.kernel;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Iterator;
-import java.util.List;
-import java.util.NoSuchElementException;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.wso2.andes.amqp.AMQPUtils;
 import org.wso2.andes.kernel.subscription.AndesSubscription;
 import org.wso2.andes.kernel.subscription.StorageQueue;
 
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.SortedMap;
+
 /**
- * Strategy definition for queue message delivery
+ * Strategy definition for queue message delivery.
  */
 public class FlowControlledQueueMessageDeliveryImpl implements MessageDeliveryStrategy {
 
@@ -45,10 +47,9 @@ public class FlowControlledQueueMessageDeliveryImpl implements MessageDeliverySt
     @Override
     public int deliverMessageToSubscriptions(StorageQueue storageQueue) throws AndesException {
 
-        Collection<DeliverableAndesMetadata> messages = storageQueue.getMessagesForDelivery();
+        SortedMap<Long, DeliverableAndesMetadata> messages =
+                (SortedMap<Long, DeliverableAndesMetadata>) storageQueue.getMessagesForDelivery();
         int sentMessageCount = 0;
-        Iterator<DeliverableAndesMetadata> iterator = messages.iterator();
-
         /*
          * get all relevant type of subscriptions.
          * For durable topic subscriptions this should return queue subscription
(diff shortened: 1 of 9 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/FlowControlledQueueMessageDeliveryImpl.java
```diff
@@ -60,38 +61,46 @@ public int deliverMessageToSubscriptions(StorageQueue storageQueue) throws Andes
...
-                    int currentConsumerIndex = consumerIndexCounter % numOfConsumers;
-                    AndesSubscription localSubscription = currentSubscriptions.get(currentConsumerIndex);
-                    consumerIndexCounter = consumerIndexCounter + 1;
-                    if (localSubscription.getSubscriberConnection().isSuspended()) {
-                        suspendedSubFound = true;
-                        continue;
-                    }
-                    if (localSubscription.getSubscriberConnection().hasRoomToAcceptMessages()) {
-
+            for (int j = 0; j < numOfConsumers; j++) {
```
pamod: let's use meaningful variable names instead of 'j'

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscription.java
```diff
@@ -590,4 +597,28 @@ public String encodeAsStr() {
...
+        this.cursor = cursor;
+    }
+
+    /**
+     * Resets the cursors of all the subscriptions bound to the queue same queue as this subscription to 0
+     */
+    private void resetCursors(){
+        List<AndesSubscription> subscriptions4Queue = storageQueue.getBoundSubscriptions();
+        for (AndesSubscription subscription : subscriptions4Queue) {
+            subscription.setCursor(0);
```
pamod: Let's avoid using magic numbers. Let's set this as a final variable

sdkottegoda:
Do we have an associated jira for this? Let's add it to the comment

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/FlowControlledQueueMessageDeliveryImpl.java
```diff
@@ -60,38 +61,46 @@ public int deliverMessageToSubscriptions(StorageQueue storageQueue) throws Andes
 
         int numOfConsumers = currentSubscriptions.size();
         int consumerIndexCounter = 0;
+        /*
+         *This is the minimum cursor that is the smallest cursor from all the bound subscriptions for a queue.
+         * All messages before this cursor is removed from the buffer.
+         */
+        long minCursor = Long.MAX_VALUE;
+        boolean isRemoved = false;
```
sdkottegoda: Shall we add a doc comment explaining the variable?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/FlowControlledQueueMessageDeliveryImpl.java
```diff
@@ -103,55 +112,58 @@ public int deliverMessageToSubscriptions(StorageQueue storageQueue) throws Andes
...
-                //if the message was delivered, it needs to be removed
-                if (numOfCurrentMsgDeliverySchedules == 1) {
-                    if (log.isDebugEnabled()) {
-                        log.debug("Removing Scheduled to send message from buffer. MsgId= " + message.getMessageID());
-                    }
-                    sentMessageCount++;
-                } else {
-
-                    //if no subscriber has a matching selector, route message to DLC queue
-                    if (!subscriberWithMatchingSelectorFound && !suspendedSubFound) {
```
sdkottegoda: If a message does not match any selector, we don't delete it now, but keep it stored since a consumer could later come without any selectors, is it?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/FlowControlledQueueMessageDeliveryImpl.java
```diff
@@ -60,38 +61,46 @@ public int deliverMessageToSubscriptions(StorageQueue storageQueue) throws Andes
...
+                if (localSubscription.getSubscriberConnection().isSuspended()) {
+                    continue;
+                }
+                Iterator<Map.Entry<Long, DeliverableAndesMetadata>> iterator = messages.tailMap(currentCursor)
+                        .entrySet().iterator();
+                while (iterator.hasNext()) {
+                    Map.Entry<Long, DeliverableAndesMetadata> entry = iterator.next();
+                    DeliverableAndesMetadata message = entry.getValue();
+                    currentCursor = entry.getKey();
+                    if (checkIfUnacked(currentSubscriptions, currentCursor)) {
```
sdkottegoda: By checking if unacked, you actually check if the message has been scheduled to be delivered to the subscriber, right? Don't you think the method name is a bit confusing?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/FlowControlledQueueMessageDeliveryImpl.java
```diff
@@ -60,38 +61,46 @@ public int deliverMessageToSubscriptions(StorageQueue storageQueue) throws Andes
...
+                        iterator.remove();
+                        isRemoved = true;
+                    } else if (localSubscription.getSubscriberConnection().hasRoomToAcceptMessages()) {
                         if (!localSubscription.getSubscriberConnection().
                                 isMessageAcceptedByConnectionSelector(message)) {
```
sdkottegoda: Can we merge this into the previous line?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/FlowControlledQueueMessageDeliveryImpl.java
```diff
@@ -60,38 +61,46 @@ public int deliverMessageToSubscriptions(StorageQueue storageQueue) throws Andes
...
-                            // If this doesn't match a selector we skip sending the message
-                            subscriberWithMatchingSelectorFound = false;
-                            continue; // continue on to match selectors of other subscribers
+                            continue; // continue on to the next message in the buffer to see if it matches
```
sdkottegoda: Don't we need to increment the cursor in this case?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/FlowControlledQueueMessageDeliveryImpl.java
riyafa: Can something like checkIfDeliveredButUnacked would do?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/FlowControlledQueueMessageDeliveryImpl.java
riyafa: It becomes longer if we do so and formatting breaks the line at message. I guess this is better than breaking the line at message

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/FlowControlledQueueMessageDeliveryImpl.java
riyafa: Cursor gets incremented at each iteration of the while loop

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/FlowControlledQueueMessageDeliveryImpl.java
riyafa: In a way Yes. We use a minCursor that is the minimum cursor of all the subscribers. Till another subscription comes all the values in front of this minCursor in the buffer would be removed but would remain in the db.

PR #909, opened 2017-06-23
Reset message cursor when last subscriber leaves

hastef88:
We deliver messages for a given queue based on a cursor that moves forward while delegating every N messages to all active subscribers. However, if the last subscriber leaves abruptly, the next subscription will not receive the messages previously allocated. Thus, resetting the cursor.

Resolves github issue : https://github.com/wso2/product-ei/issues/689


Asitha:
Merged the changes to 4.x.x branch with PR https://github.com/wso2/andes/pull/941

PR #911, opened 2017-06-30
Changing MQTT message router value

Supun94:
Display MQTT related topics, subscription and message details. Implemented MQTT Admin service and service stub

### DIFF
--- a/modules/andes-core/broker-plugins/firewall/MANIFEST.MF
+++ b/modules/andes-core/broker-plugins/firewall/MANIFEST.MF
@@ -31,6 +31,10 @@ Import-Package:  org.apache.qpid,
  javax.management;version=1.0.0,
  javax.management.openmbean;version=1.0.0,
  org.osgi.util.tracker;version=1.0.0,
- org.osgi.framework;version=1.3
+ org.osgi.framework;version=1.3,
+ com.googlecode.cqengine,
+ org.wso2.carbon.metrics.manager,
+ org.wso2.carbon.metrics.manager,
+ io.netty.buffer
 Private-Package: org.wso2.andes.server.security.access.config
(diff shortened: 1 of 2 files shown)


modules/andes-core/broker-plugins/firewall/MANIFEST.MF
```diff
@@ -31,6 +31,10 @@ Import-Package:  org.apache.qpid,
  javax.management;version=1.0.0,
  javax.management.openmbean;version=1.0.0,
  org.osgi.util.tracker;version=1.0.0,
- org.osgi.framework;version=1.3
+ org.osgi.framework;version=1.3,
+ com.googlecode.cqengine,
```
madhawa-gunasekara: Shall we specify the version ranges here?

madhawa-gunasekara:
Merged. Thanks for the contribution Supun.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscriptionManager.java
```diff
@@ -459,6 +459,11 @@ private void removeRemoteSubscriptionFromMemory(AndesSubscription subscription)
 
         Set<AndesSubscription> filteredSubscriptions = new HashSet<>();
         String messageRouter = destinationType.getAndesMessageRouter();
+
+        if (ProtocolType.MQTT.equals(protocolType)) {
```
hastef88: This is done because we do not have a mqtt_topic / jms_topic separation when querying at the database level. 

Introducing such a separation at the DestinationType Enum would incur a lot of mqtt-specific code, and given that we are not supporting mqtt in the future releases, we are resorting to this approach.

PR #693, opened 2016-09-06
Add a new method for filtering subscriptions for UI.

abeykoon:

PR #912, opened 2017-07-03
Fix #693 of EI, the memory leak caused by charbuf threadlocal variable

Manuri:
This PR fixes https://github.com/wso2/product-ei/issues/693 which is caused by the charbuf theradlocal variable of org.wso2.andes.util.Strings class

### DIFF
--- a/modules/andes-core/common/src/main/java/org/wso2/andes/util/Strings.java
+++ b/modules/andes-core/common/src/main/java/org/wso2/andes/util/Strings.java
@@ -22,6 +22,7 @@ package org.wso2.andes.util;
 
 import java.io.UnsupportedEncodingException;
 
+import java.nio.charset.Charset;
 import java.util.Arrays;
 import java.util.Map;
 import java.util.Properties;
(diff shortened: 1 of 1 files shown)


modules/andes-core/common/src/main/java/org/wso2/andes/util/Strings.java
```diff
@@ -40,13 +41,10 @@
...
-    {
-        public char[] initialValue()
-        {
-            return new char[4096];
-        }
-    };
+    /**
+     * Resolving the UTF_8 charset
+     */
+    private static final Charset UTF_8_CHARSET = Charset.forName("UTF-8");
```
ramith: have you seen this one : http://docs.oracle.com/javase/7/docs/api/java/nio/charset/StandardCharsets.html#UTF_8 ?

Manuri: Yes I saw this. I had in mind that we need to support JDK6. Can use this for the public branch since theres no need to support JDK 6. Thanks for pointing this out.

PR #916, opened 2017-08-23
Remove Hazelcast from Cluster manager

a5anka:
Following changes were done to remove Hazelcast from cluster manager

-  Hazelcast based network partition listener was removed since it is not required without Hazelcast
- Hazlecast based unique node ID generator was removed since it is not required for a 2 node HA (active-passive) setup
- Removed the Hazelcast map used to detect duplicate node IDs
- Removed Hazelcast socket information from node details


modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/ClusterAgent.java
```diff
@@ -70,20 +61,10 @@
      * @param manager
      *         Cluster manager for the current node
      */
-    void start(ClusterManager manager) throws AndesException;
+    String start(ClusterManager manager) throws AndesException;
```
abeykoon: What do we return out of this method?

a5anka: Current node's registered node ID in the cluster.

abeykoon:
How node IDs are generated/picked up in the new coordination mechanism? Are they removed or are they picked up by config files?

a5anka:
It is a bit similar to what we had before. If a node id is provided in the config that will be used. Otherwise, a unique ID will be generated.

PR #921, opened 2017-09-14
Active Passive HA implementation

a5anka:


modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
```diff
@@ -160,7 +160,7 @@ public static void messageReceived(IncomingMessage incomingMessage, AndesChannel
      * @param andesMessage Message coming in
      * @param andesChannel AndesChannel
      */
-    public static void messageReceived(AndesMessage andesMessage, AndesChannel andesChannel) {
+    public static void messageReceived(AndesMessage andesMessage, AndesChannel andesChannel) throws AndesException {
```
vinok88: Update the doc comment as well with throws part

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/dtx/DistributedTransaction.java
```diff
@@ -42,6 +44,11 @@
 public class DistributedTransaction {
 
     /**
+     * Class logger
+     */
+    private static final Log LOGGER = LogFactory.getLog(DistributedTransaction.class);
```
hastef88: Better stick to a single naming convention for loggers ? lower case is the most common I see.

modules/andes-core/broker/src/main/java/org/wso2/andes/server/handler/ChannelFlowHandler.java
```diff
@@ -58,29 +59,35 @@ public void methodReceived(AMQStateManager stateManager, final ChannelFlowBody b
...
-                session.writeFrame(channelException.getCloseFrame(channelId));
-                try {
-                    session.closeChannel(channelId);
-                } catch (AMQException e) {
-                    logger.error("Couldn't close channel (channel id ) " + channelId, e);
                 }
+            });
+        } catch (AndesException e) {
+            throw new AMQException(AMQConstant.INTERNAL_ERROR,
+                                   "Error occurred while while handling flow command.", e);
```
hastef88: typo. double "while"

modules/andes-core/broker/src/main/java/org/wso2/andes/server/handler/ChannelFlowHandler.java
```diff
@@ -58,29 +59,35 @@ public void methodReceived(AMQStateManager stateManager, final ChannelFlowBody b
...
-                AMQChannelException channelException = body.getChannelException(AMQConstant.INTERNAL_ERROR,
-                                                                                exception.getMessage());
-                session.writeFrame(channelException.getCloseFrame(channelId));
-                try {
-                    session.closeChannel(channelId);
-                } catch (AMQException e) {
-                    logger.error("Couldn't close channel (channel id ) " + channelId, e);
                 }
+            });
+        } catch (AndesException e) {
```
sdkottegoda: Shouldn't this go inside onException? I mean if an exception is thrown while QpidAndesBridge.notifyChannelFlow() is executed, don't we need to send a channel close frame?

modules/andes-core/broker/src/main/java/org/wso2/andes/mqtt/connectors/PersistenceStoreConnector.java
```diff
@@ -141,7 +141,11 @@ public void addMessage(MQTTMessageContext messageContext) throws MQTTException {
             // Publish to Andes core
             AndesMessage andesMessage = new MQTTMessage(messageHeader);
             andesMessage.addMessagePart(messagePart);
-            Andes.getInstance().messageReceived(andesMessage, publisher.getChannel(), messageContext.getPubAckHandler());
+            try {
+                Andes.getInstance().messageReceived(andesMessage, publisher.getChannel(), messageContext.getPubAckHandler());
+            } catch (AndesException e) {
+                throw new MQTTException("Error while publishing messages", e);
```
vinok88: shall we give a more specific message?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/dtx/DistributedTransaction.java
a5anka: I think this is the correct one since this is a `static final`.

modules/andes-core/broker/src/main/java/org/wso2/andes/server/handler/ChannelFlowHandler.java
a5anka: OnException is triggered if you get an error while processing the event. This catch statement is to catch exceptions thrown while publishing the event. Additionally, when the exception is received the protocol engine will initiate the channel close. So we do not have to do anything.

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/QpidAndesBridge.java
a5anka: :+1:

modules/andes-core/broker/src/main/java/org/wso2/andes/server/handler/ChannelFlowHandler.java
a5anka: :+1: Fixed

modules/andes-core/broker/src/main/java/org/wso2/andes/mqtt/connectors/PersistenceStoreConnector.java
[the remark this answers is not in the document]
a5anka: MQTT is not supported in EI7. Therefore will not add much value. We are anyway going to get rid of this code.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/disruptor/inbound/InboundQueueEvent.java
```diff
@@ -159,6 +159,11 @@ public String eventInfo() {
         return eventType.toString();
     }
 
+    @Override
+    public boolean isActionableWhenPassive() {
```
sdkottegoda: What will happen when a subcription comes binding to a new queue name? The queue event will be taken in and the queue will get created, and the binding will not be taken in. Will that be handled properly when the subscription tries to get subscribed from the other node?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/dtx/DistributedTransaction.java
```diff
@@ -253,7 +260,11 @@ public void enqueueMessage(AndesMessage andesMessage, AndesChannel andesChannel)
                 }
             }
         } else {
-            QpidAndesBridge.messageReceived(andesMessage, andesChannel);
+            try {
+                QpidAndesBridge.messageReceived(andesMessage, andesChannel);
+            } catch (AndesException e) {
+                LOGGER.error("Non-transactional message publishing failed", e);
```
sdkottegoda: Is it correct to say "Non-transactional message publishing failed"?

modules/andes-core/broker/src/main/java/org/wso2/andes/server/connection/ConnectionRegistry.java
```diff
@@ -38,22 +38,28 @@ public void initialise()
...
     }
 
-    /** Close all of the currently open connections. */
-    public void close()
-    {
+    /**
+     * Close all existing client connections.
+     */
+    public void close() {
+        close("Broker is shutting down");
```
sdkottegoda: we need to come up with a better message, right?

a5anka: Yes. The message we currently does not match the real reason. Will do the required change in a future PR.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/dtx/DistributedTransaction.java
a5anka: Yes. If we do a message publishing without sending a dtx.start, it is non-transational.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/disruptor/inbound/InboundQueueEvent.java
a5anka: Can you suggest a test scenario to verify this?

PR #922, opened 2017-09-18
Adding authentication interface support for C5

himasha:


modules/orbit/andes/pom.xml
```diff
@@ -145,6 +145,7 @@
             com.google.common.base;version="19.0.0",
             com.google.common.util.concurrent;version="19.0.0",
             com.google.common.cache;version="19.0.0",
+            org.wso2.carbon.business.messaging.core.authentication,
```
a5anka: Even though we add the  CBM here, I don't see it used in this PR. Why is it?

himasha: Since we moved CarbonBasedPrincipalDatabase class to carbon-business-messaging component, we need to access it when calling 'passwordcb.isAuthenticated()' in PlainSaslServer class.

PR #924, opened 2017-09-20
Add YAML based configuration support.

vinok88:
Add carbon-config dependancy and the plugin for config doc generation.
Add Configuration service singleton to fetch the root config instance.
Add model classes for the configurations in broker.xml
Use BrokerConfigurationService.getInstance().getBrokerConfiguration() to
fetch a config value.
NOTE : This commit only add support for yaml based configs, conversion
of broker.xml to deployment.yaml is not done.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/BrokerConfigurationService.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/BrokerConfigurationService.java
@@ -32,8 +32,17 @@ import java.nio.file.Paths;
 public class BrokerConfigurationService {
     private static Log log = LogFactory.getLog(BrokerConfigurationService.class);
 
+    /**
+     * Configuration directory name.
+     */
     private static String CONF_FOLDER = "conf";
+    /**
+     * Carbon runtime name.
+     */
     private static String RUNTIME = "broker";
+    /**
+     * Deployment descriptor name.
+     */
     private static String CONFIG_NAME = "deployment.yaml";
     private static BrokerConfigurationService brokerConfigurationService = new BrokerConfigurationService();
     private BrokerConfiguration brokerConfiguration;
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/BrokerConfigurationService.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/BrokerConfigurationService.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 Inc. licenses this file to you under the Apache License, 
+ * Version 2.0 (the "License"); you may not use this file except 
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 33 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/models/BrokerConfiguration.java
```diff
@@ -0,0 +1,40 @@
...
+ * This class provide access to all the configuration object representations of the broker.
+ */
+@Configuration(namespace = "wso2.carbon.broker",
+               description = "Broker configuration parameters")
+public class BrokerConfiguration {
+
+    @Element(description = "Coordination related configuration")
+    private CoordinationConfiguration coordination = new CoordinationConfiguration();
+
+    public CoordinationConfiguration getCoordination() {
```
a5anka: Do we need to provide getters and setters? Aren't they generated from `@Element` annotation?

vinok88: According to the sample in carbon-config, it is required.

a5anka: :+1:

vinok88: I checked back again. We only need getter methods. Setters not required. My bad. Thanks for pointing. https://github.com/wso2/carbon-config/blob/config-docs-plugin/samples/config-generator/src/main/java/org/wso2/carbon/config/samples/ParentConfiguration.java

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/BrokerConfigurationService.java
```diff
@@ -0,0 +1,71 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+/**
+ * Broker configuration provider.
+ */
+public class BrokerConfigurationService {
+    private static Log log = LogFactory.getLog(BrokerConfigurationService.class);
+
+    private static String CONF_FOLDER = "conf";
```
ramith: java docs for these variables ( as a practice would be nice to have).

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/models/CoordinationConfiguration.java
```diff
@@ -0,0 +1,41 @@
...
+import org.wso2.carbon.config.annotation.Element;
+
+/**
+ * Configuration model for broker coordination.
+ */
+@Configuration(description = "Broker coordination config")
+public class CoordinationConfiguration {
+
+    @Element(description = "The node ID of each member")
+    private String nodeID = "default";
```
ramith: whats the convention for id is it {Id, ID} ?

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/models/CoordinationConfiguration.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Element(description = "Thrift is used to maintain and sync slot (message groups) ranges between MB nodes.")
+    private String thriftServerHost = "localhost";
```
ramith: suggestion, shall we rename the 'thrift' part of this functionality in such way it describes what the functionality instead of highlighting the name of the transport protocol used ( i.e thrift) @Asitha @indikasampath2000 @a5anka @himasha @nandika @pamod

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/models/CoordinationConfiguration.java
vinok88: We need to have whatever the name that should be in yaml config. This is what was in broker.xml

PR #925, opened 2017-09-20
Authentication interface support for C5. Also includes the configurat…

himasha:
…ion required to define authentication implementation.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
@@ -83,6 +83,13 @@ public enum AndesConfiguration implements ConfigurationProperty {
     COORDINATION_NODE_ID("coordination/nodeID", "default", String.class),
 
     /**
+     * Class name of the authentication interface implementation is added here.By default,default implementation of
+     * Broker component is added.
+     */
+    AUTHENTICATOR_CLASS("authenticator/@class",
+            "org.wso2.carbon.business.messaging.core.internal.AuthenticationServiceImpl", String.class),
+
+    /**
      * The IP address to which mqtt/amqp channels should be bound.
      */
     TRANSPORTS_BIND_ADDRESS("transports/bindAddress", "*", String.class),
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
@@ -46,25 +46,24 @@ public enum AndesConfiguration implements ConfigurationProperty {
      * Setting this to a very low value could have an impact on the performance since the number of DB calls will be
      * high. Setting this to a large value increases the time taken for a cluster event to be synchronized.
      */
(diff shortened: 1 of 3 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
```diff
@@ -46,25 +46,24 @@
      * Setting this to a very low value could have an impact on the performance since the number of DB calls will be
```
vinok88: Looks like you have formatted the whole class. This makes hard to identify the change done. can you include only your change? (I assume that the existing class do not have a formatting issue)

himasha: The existing class already has formatting issues. To address your point, will adding a separate commit for the new property make sense?

modules/orbit/andes/pom.xml
```diff
@@ -145,6 +145,7 @@
             com.google.common.base;version="19.0.0",
             com.google.common.util.concurrent;version="19.0.0",
             com.google.common.cache;version="19.0.0",
+            org.wso2.carbon.business.messaging.core.authentication,
```
hemikak: A bit confused here. cbm is a dependency of andes as I know. So when releasing andes we to build cbm ? Or am I wrong ? :)

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
a5anka: It seems some of the formatting changes are not required. Are you using the right margin as 120?

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
```diff
@@ -210,42 +212,44 @@
...
+    LIST_TRANSPORT_MQTT_AUTHENTICATION_PROPERTIES("transports/mqtt/security/authenticator/property/@name", "",
+            List.class),
 
     /**
      * This can be used to access a property by giving its key. e.g. hosturl
      */
-    TRANSPORT_MQTT_AUTHENTICATION_PROPERTIES("transports/mqtt/security/authenticator/property[@name = '{key}']", "", String.class),
-    /**
+    TRANSPORT_MQTT_AUTHENTICATION_PROPERTIES("transports/mqtt/security/authenticator/property[@name = '{key}']", "",
+            String.class), /**
```
a5anka: Shall we fix the formatting issue? Doc comment should start in a new line.

modules/orbit/andes/pom.xml
a5anka: Yes, It's a circular dependency. Andes need CBM and CBM needs Andes. Something that needs to be fixed before release.

modules/andes-core/broker/src/main/java/org/wso2/andes/configuration/enums/AndesConfiguration.java
himasha: Reverted back to previous formatting and included the new property with 0450a50

PR #936, opened 2017-11-02
Fix NullPointerException during messaging

sdkottegoda:
## Purpose
Resolves wso2/product-ei#1113

## Goals
Clears references for objects within Disruptor after usage.

## Approach
Last event handler of the Disruptor clears all the object reference from the Disruptor event container objects. In the outbound disruptor, we have missed clearing some newly introduced objects, With this fix, we have added a new handler for clearing those objects as well from the Disruptor buffer.

## User stories
N/A, this is a bug fix to prevent the server from going OOM.

## Release note
Clear references for objects within Disruptor after usage.

## Documentation
N/A, this is a bug fix

## Training
N/A, this is a bug fix

## Certification
N/A, this is a bug fix

## Marketing
N/A, this is a bug fix

## Automation tests
N/A, Need to manually test a heap dump to identify this. Since this is a probable OOM scenario.

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes
 - Ran FindSecurityBugs plugin and verified report? yes
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes

## Samples
N/A, this is a bug fix

## Related PRs
N/A, this is a bug fix

## Migrations (if applicable)
N/A, this is a bug fix

## Test environment
This can be reproduced with any JDK, OS and DB
 
## Learning
This issue can be identified by taking a heap dump after publishing and consuming the message from the broker and checking for references of ProtocolMessage, AndesChannel, LocalSubscription objects from Disruptor.

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/disruptor/delivery/DeliveryEventCleanupHandler.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/disruptor/delivery/DeliveryEventCleanupHandler.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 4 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/disruptor/delivery/DeliveryEventCleanupHandler.java
```diff
@@ -0,0 +1,14 @@
+package org.wso2.andes.kernel.disruptor.delivery;
```
chanikag: Missing license header?

PR #948, opened 2018-01-04
Fix NPE when subscriber closed and Thrift server thread hang and Thrift server hangs in socket read

wggihan:
## Purpose
To fix https://github.com/wso2/product-ei/issues/1619 and https://github.com/wso2/product-ei/issues/1613. 

## Goals
Fixes the NPE that occurs when a subscription connection is abruptly removed. The issue was caused by a message duplication. Messages that are failed during delivery were requeued where all the messages that are bound to the subscription are also requeued from another path. As the resolution, avoided requeueing messages when the delivery is failed due to the subscriber being not available.

This fix provided by @sdkottegoda .

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPLocalSubscription.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPLocalSubscription.java
@@ -29,6 +29,7 @@ import org.wso2.andes.kernel.AndesMessageMetadata;
 import org.wso2.andes.kernel.ProtocolDeliveryFailureException;
 import org.wso2.andes.kernel.ProtocolDeliveryRulesFailureException;
 import org.wso2.andes.kernel.ProtocolMessage;
+import org.wso2.andes.kernel.SubscriptionAlreadyClosedException;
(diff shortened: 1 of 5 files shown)


sdkottegoda:
The 2 commits fix the 2 issues separetely. Shall we correct the commit messages accordingly?

modules/andes-core/broker/src/main/java/org/wso2/andes/amqp/AMQPLocalSubscription.java
```diff
@@ -261,6 +264,13 @@ private void sendMessage(QueueEntry queueEntry) throws AndesException {
                 throw new AndesException("Error occurred while delivering message. Unexpected Subscription type for "
                         + "message with ID : " + messageID);
             }
+        } catch (AMQSubscriptionClosedException e) {
+            ProtocolMessage protocolMessage = ((AMQMessage) queueEntry.getMessage()).getAndesMetadataReference();
+            log.error("AMQP Protocol Error while delivering message to the subscriber subID= "
```
madhawa-gunasekara: Shall we fix this? log and throw is anti-pattern

PR #956, opened 2018-01-30
Fix product-ei/issues/1716

GDLMadushanka:
## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes
 - Ran FindSecurityBugs plugin and verified report? yes
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/SlotManagementServiceImpl.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/SlotManagementServiceImpl.java
@@ -87,6 +87,17 @@ public class SlotManagementServiceImpl implements SlotManagementService.Iface {
         }
     }
 
+    /**
+     * Upon receiving the health-check call return true
+     *
+     * @return always true
+     * @throws TException thrift exception
+     */
+    @Override
+    public boolean healthCheck() throws TException {
+        return true;
+    }
+
     @Override
     public void reAssignSlotWhenNoSubscribers(String nodeId, String queueName) throws TException {
         if (AndesContext.getInstance().getClusterAgent().isCoordinator()) {
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/ThriftClientFactory.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/ThriftClientFactory.java
@@ -23,6 +23,7 @@ import org.apache.commons.logging.LogFactory;
 import org.apache.commons.pool2.BasePooledObjectFactory;
 import org.apache.commons.pool2.PooledObject;
 import org.apache.commons.pool2.impl.DefaultPooledObject;
+import org.apache.thrift.TException;
 import org.apache.thrift.protocol.TBinaryProtocol;
 import org.apache.thrift.protocol.TProtocol;
 import org.apache.thrift.transport.TSocket;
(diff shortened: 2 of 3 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/slot/gen/SlotManagementService.java
```diff
@@ -154,6 +158,29 @@ public boolean recv_deleteSlot() throws org.apache.thrift.TException
...
+    }
+
+    public boolean recv_healthCheck() throws org.apache.thrift.TException
+    {
+      healthCheck_result result = new healthCheck_result();
+      receiveBase(result, "healthCheck");
+      if (result.isSetSuccess()) {
+        return result.success;
+      }
+      throw new org.apache.thrift.TApplicationException(org.apache.thrift.TApplicationException.MISSING_RESULT, "healthCheck failed: unknown result");
```
Bhashinee: Exceeds 120.

modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/ThriftClientFactory.java
```diff
@@ -96,4 +98,22 @@ public void destroyObject(PooledObject<SlotManagementService.Client> client) thr
...
+     *
+     * @param client client
+     * @return connection's usability
+     */
+    @Override
+    public boolean validateObject(PooledObject<SlotManagementService.Client> client) {
+        try {
+            return client.getObject().healthCheck(Constants.HEALTH_CHECK);
+        } catch (TException e) {
+            log.error("exception occurs when validating the thrift connection " + e);
```
chanikag: It would be better to pass the exception rather than append. i.e,
 log.error("exception occurs when validating the thrift connection ", e);
WDYT?

modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/SlotManagementServiceImpl.java
```diff
@@ -87,6 +88,21 @@ public boolean deleteSlot(String queueName, SlotInfo slotInfo, String nodeId) th
...
+    /**
+     * Used to check the health of the connection taken from thrift connection pool.
+     *
+     * @param healthCheckStr The string received to healthCheck.
+     * @return received string is the expected or not.
+     * @throws TException thrift exception.
+     */
+    @Override
+    public boolean healthCheck(String healthCheckStr) throws TException {
+        if (healthCheckStr != null && healthCheckStr.equals(Constants.HEALTH_CHECK)) {
```
nadeeshaan: Instead of using the if else clause, what if we return the condition directly? WDYT?

modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/ThriftClientFactory.java
GDLMadushanka: +1 changed

modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/SlotManagementServiceImpl.java
GDLMadushanka: implementation is now changed

PR #960, opened 2018-02-23
Fix wso2/product-ei/issues/1883

GDLMadushanka:

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/ThriftClientFactory.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/ThriftClientFactory.java
@@ -111,7 +111,7 @@ public class ThriftClientFactory extends BasePooledObjectFactory<SlotManagementS
         try {
             return client.getObject().healthCheck();
         } catch (TException e) {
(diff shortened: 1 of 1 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/thrift/ThriftClientFactory.java
```diff
@@ -111,7 +111,7 @@ public boolean validateObject(PooledObject<SlotManagementService.Client> client)
         try {
             return client.getObject().healthCheck();
         } catch (TException e) {
-            log.error("exception occurs while validating the thrift connection" + e);
+            log.debug("exception occurs while validating the thrift connection" + e);
```
nuwanw: check whether debug is enabled before logging, It will optimise code

nuwanw: parse the e as argument rather than appending to message

nuwanw: seems the error message is misleading because it says exception occurs while validating. Lets change it to "Connection is not active. Hence removing the connection from pool"

PR #978, opened 2018-06-14
Fix Issue in Active Durable Subscription Search

sdkottegoda:
## Purpose
Resolves https://github.com/wso2/product-ei/issues/2320

## Approach
Fixes the issue of active durable subscriptions not being listed once searched by the subscription identifier, by having the filtration done using the storage queue associated with the subscription as well as the subscription id.


vinok88:
Issue id missing in the commit message.

PR #999, opened 2018-09-17
Commit on select statements

sdkottegoda:
## Purpose
Resolves https://github.com/wso2/product-ei/issues/2672


a5anka:
Shall we add the commits to RDBMSDtxStoreImpl class as well?

PR #1011, opened 2019-03-29
Remove 'carbon:' from durable topic subscription mbean name

ThishaniLucas:

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/registry/StorageQueueRegistry.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/registry/StorageQueueRegistry.java
@@ -161,9 +161,10 @@ public class StorageQueueRegistry {
     }
 
     private Hashtable<String, String> getMbeanObjectnameProp(String storageQueue) {
+        int index = storageQueue.indexOf(':');
         Hashtable<String, String> tab = new Hashtable<>();
         tab.put("type", TOPIC_INFO_MBEAN_TYPE);
-        tab.put("name", ObjectName.quote(storageQueue));
+        tab.put("name", ObjectName.quote(storageQueue.substring(index + 1)));
         return tab;
     }
 }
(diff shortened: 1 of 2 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/registry/StorageQueueRegistry.java
```diff
@@ -161,9 +161,10 @@ private boolean isMbeanRegistered(String storageQueue) {
     }
 
     private Hashtable<String, String> getMbeanObjectnameProp(String storageQueue) {
+        int index = storageQueue.indexOf(':');
```
sdkottegoda: Shall we look for "carbon:" here. It is possible to declare queues with ':' upon changing a configuration. Also we can use DURABLE_SUBSCRIPTION_QUEUE_PREFIX in AndesConstants for this purpose?

modules/andes-core/broker/src/main/java/org/wso2/andes/server/information/management/DurableTopicSubscriptionInformationMBean.java
```diff
@@ -47,6 +47,7 @@ public long getMsgCount() {
 
     @Override
     public String getObjectInstanceName() {
-        return ObjectName.quote(storageQueue);
+        int index = storageQueue.indexOf(':');
```
sdkottegoda: same as above

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/registry/StorageQueueRegistry.java
ThishaniLucas: indexOf will anyway return the first occurrence of the char ':'. Therefore, I think it's safe to use because a storage queue name will always start with 'carbon:'. If you still want to check for 'carbon:', I shall change it.

sdkottegoda: Is this method called only for durable topic subscriptions? Isn't it called for normal queues? Then I suppose we can go ahead with the colon. But this will be invalid if for some reason we change the DURABLE_SUBSCRIPTION_QUEUE_PREFIX. What do you think we should do?

ThishaniLucas: Please correct me if I'm wrong, but I don't think it will happen. 'carbon' is the default virtual host name we use and always append the virtual host name to subscription ID to differentiate the internal queue of a durable subscription. We don't support any other virtual host name. If we are to change this prefix, IMHO it will have effect in many other places. Therefore, I suppose this should be fine. Also, yes, this method is called only in the flow of durable subscriptions.

PR #1016, opened 2019-04-13
Update branch with the master branch

malakaganga:
## Purpose
Update branch with the master branch


a5anka:
@malakaganga @indikasampath2000 when we pull changes like this from master the git history is lost. Ideally, we should have merged the commits to this branch from master. Shall we do it like that the next time?

PR #1023, opened 2019-07-08
Fix message broker unstable with subscription issue

dulanjalidilmi:
Fix wso2/product-ei#4336
Fix wso2/product-ei#4337

## Approach

Unable to create a durable topic subscription with the same name when the servers crash
This was fixed by removing all durable and non-durable subscriptions in cluster initialization.

NullPointerException in MB clusters when there are network delays
This was fixed by establishing the same old binding when a network delay occurs.


dulanjalidilmi:
Closing this PR because above issues is resolved from https://github.com/wso2/andes/pull/1040

PR #1037, opened 2019-11-23
[Fix EI-4795] MB XDeliveryCount header does not reflect correct values

abeykoon:
## Purpose

[Fix EI-4795] MB XDeliveryCount header does not reflect correct values

## Goals

Fixes: https://github.com/wso2/product-ei/issues/4795

## Approach

Fix the race condition on requeing messages 

## Documentation

N/A

## Training

N/A

## Certification

N/A

## Marketing

N/A

## Automation tests

Will add integration test 

## Security checks

yes

### DIFF
--- a/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/DeliverableAndesMetadata.java
+++ b/modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/DeliverableAndesMetadata.java
@@ -251,10 +251,14 @@ public class DeliverableAndesMetadata extends AndesMessageMetadata {
         channelInformation.addChannelStatus(ChannelMessageStatus.DISPATCHED);
 
         if (!this.isBeyondLastRollbackedMessage) {
-            channelInformation.incrementDeliveryCount();
+            int i = channelInformation.incrementDeliveryCount();
+            MessageTracer.trace(getMessageID(), getDestination(), "Message within rollback id. "
+                    + "Current count " + i);
         } else {
             // No need to increase deliveryCount if this message is beyond the last rollback.
             MessageTracer.trace(getMessageID(), getDestination(), MessageTracer.MESSAGE_BEYOND_LAST_ROLLBACK);
+            MessageTracer.trace(getMessageID(), getDestination(), MessageTracer.MESSAGE_BEYOND_LAST_ROLLBACK
+                    + " current count = " + channelInformation.getDeliveryCount());
         }
     }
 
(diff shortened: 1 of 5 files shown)


modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/DeliverableAndesMetadata.java
```diff
@@ -251,10 +251,14 @@ public void markAsDispatchedToDeliver(UUID channelID) {
         channelInformation.addChannelStatus(ChannelMessageStatus.DISPATCHED);
 
         if (!this.isBeyondLastRollbackedMessage) {
-            channelInformation.incrementDeliveryCount();
+            int i = channelInformation.incrementDeliveryCount();
+            MessageTracer.trace(getMessageID(), getDestination(), "Message within rollback id. "
+                    + "Increased count " + i);
```
sdkottegoda: I suppose increased count is a bit misleading. We can use the term current count as used in the else block

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscription.java
```diff
@@ -248,12 +248,14 @@ public void onMessageAck(long messageID) throws AndesException {
...
+     * @param isMessageBeyondLastRollback true if message was rejected after last rollback
      * @throws AndesException on a message re-schedule issue
      * @return DeliverableAndesMetadata reference of rejected message
      */
-    public DeliverableAndesMetadata onMessageReject(long messageID, boolean reQueue)
-            throws AndesException {
+    public DeliverableAndesMetadata onMessageReject(long messageID, boolean reQueue,
+                                                    boolean isMessageBeyondLastRollback) throws AndesException {
         DeliverableAndesMetadata rejectedMessage = subscriberConnection.onMessageReject(messageID);
+        rejectedMessage.setIsBeyondLastRollbackedMessage(isMessageBeyondLastRollback);
```
sdkottegoda: Should this be done before subscriberConnection.onMessageReject?

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/DeliverableAndesMetadata.java
abeykoon: Fixed.

modules/andes-core/broker/src/main/java/org/wso2/andes/kernel/subscription/AndesSubscription.java
abeykoon: What on message reject does is removing the message from outbound tracker and getting the message reference at the server side. We need to call setIsBeyondLastRollbackedMessage on that message.

PR #1054, opened 2022-06-24
Remove the hazelcast dependency from andes

1akshitha:
As we have deprecated the hazelcast based distributed caching use cases in APIM, removing the hazelcast dependency from andes repo.


modules/andes-core/broker/src/main/java/org/wso2/andes/server/cluster/coordination/CoordinationComponentFactory.java
```diff
@@ -79,25 +70,11 @@ public ClusterNotificationListenerManager createClusterNotificationListener() th
     public ClusterNotificationAgent createClusterNotificationAgent() throws AndesException {
         initConfiguration();
         AndesContextStore contextStore = AndesContext.getInstance().getAndesContextStore();
-        ClusterNotificationAgent clusterNotificationAgent;
+        ClusterNotificationAgent clusterNotificationAgent = new StandaloneMockNotificationAgent();
         if (isClusteringEnabled) {
             if (isRDBMSBasedCoordinationEnabled) {
```
Nirothipan, line 75: we can merge these if conditions.

Nirothipan (approved):

